PR #8, opened 2019-03-14
Add scripts to run tests

bratseth:
@toregge please review

### DIFF
--- a/lib/test_base.rb
+++ b/lib/test_base.rb
@@ -60,43 +60,37 @@ module TestBase
   DRUBY_REMOTE_PORT = 27183
 
   # absolute path to the systemtests/tests/ directory as specified by RUBYLIB
-  TESTDIR = $:.find {|path| path =~ /systemtests.*\/tests/}
+  OLDTESTDIR = $:.find {|path| path =~ /systemtests.*\/tests/}
+  if OLDTESTDIR == nil
+    OLDTESTDIR = ""
+  end
+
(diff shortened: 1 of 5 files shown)


lib/test_base.rb
```diff
@@ -60,43 +60,37 @@ module TestBase
   DRUBY_REMOTE_PORT = 27183
 
   # absolute path to the systemtests/tests/ directory as specified by RUBYLIB
-  TESTDIR = $:.find {|path| path =~ /systemtests.*\/tests/}
+  OLDTESTDIR = $:.find {|path| path =~ /systemtests.*\/tests/}
+  if OLDTESTDIR == nil
+    OLDTESTDIR = ""
```
toregge: This is trying to change a constant. Changing the code to

OLDTESTDIR = $:.find {|path| path =~ /systemtests.*\/tests/} || ""

would eliminate the issue.  A better approach is to add a function to EnvironmentBase that returns the string.

lib/test_base.rb
```diff
@@ -60,43 +60,37 @@ module TestBase
...
+  end
+
+  # absolute path to the system-test/tests/ directory as specified by RUBYLIB                                                                                
+  TESTDIR = $:.find {|path| path =~ /system-test.*\/tests/}
   if TESTDIR == nil
     TESTDIR = ""
```
toregge: This is also problematic, TESTDIR is a constant.

bratseth:
I just realized that this wouldn't work since we install to "systemtests" in both cases. This should work though.

PR #31, opened 2019-03-21
Update test to match actual result. MERGEOK

aressem:


aressem:
Needed to get the other system test fixes built and tested.

PR #51, opened 2019-04-05
Ensure that we do not get random filename if there are more than one …

baldersheim:
…file.
@hmusum PR


baldersheim:
@lesters is probably at work today.

PR #68, opened 2019-04-27
Revert "Ensure that docs are fed in order."

baldersheim:
Reverts vespa-engine/system-test#67
@hmusum or @geirst or @toregge PR Too slow


jonmv:
This leaves the original test broken, and blocks release.

PR #69, opened 2019-04-28
Preserve feed order when verifying metrics.

baldersheim:
@geirst or @toregge or @hmusum PR


jonmv:
Looks good, but is it used?

PR #77, opened 2019-05-02
Use correct NodeFlavors

freva:
Same as vespa/systemtests/pull/4359

Might need an export package for `com.yahoo.config.provision.internal`?


hakonhall:
No, that should be fine.

PR #79, opened 2019-05-02
Only add vespafeed tag for xml data

baldersheim:
@hmusum or @aressem PR


baldersheim:
I am not sure. I will see what happens and update if so.

PR #88, opened 2019-05-06
Revert "Try to disable test."

aressem:
Reverts vespa-engine/system-test#87

Just review. I will observe the tests before reverting.


aressem:
Did not have any positive effect, so reverting this.

PR #96, opened 2019-05-13
Use /bin/lz4 directly in core_pattern.

aressem:
Realized that this is broken anyway because the vespa-core-dumper is not open sourced.


aressem:
Do not merge this one yet. Just review.

PR #100, opened 2019-05-15
Fix reference to logserver port (fixes bug in https://github.com/vesp… MERGEOK

hmusum:
…a-engine/system-test/pull/98)


aressem:
Merge ok since this fixes broken system tests.

PR #108, opened 2019-05-20
Allow finer control of numthreads for feeder.

baldersheim:
@hmusum or @aressem PR


baldersheim:
Just approve. I will coordinate merge.

PR #137, opened 2019-06-06
add reproduction for bug when partial update of both map and ngram fi..

arnej27959:
…eld happens

@geirst please review and merge
@bratseth @vekterli FYI


tests/search/bugs/map-and-ngram/bug_map_and_ngram.rb
```diff
@@ -0,0 +1,35 @@
...
+               search(SearchCluster.new("multitest").
+                      num_parts(1).redundancy(1).ready_copies(1).
+                      sd(selfdir+"foo.sd").
+                      indexing("mycc")))
+    start
+    feed_and_wait_for_docs("foo", 3, :file => selfdir+"feed.xml")
+
+    node = vespa.adminserver
+    node.copy(selfdir + "updates.json", dirs.tmpdir)
+    node.execute("cd #{dirs.tmpdir} && vespa-feeder --trace 4 --maxpending 1 updates.json");
```
geirst: Please use the feed() function instead. Copying files is done automatically. Extra params can be specified as well. See here for which params are supported: https://github.com/vespa-engine/system-test/blob/master/lib/nodetypes/feeder.rb#L147.

tests/search/bugs/map-and-ngram/feed.xml
```diff
@@ -0,0 +1,29 @@
+<document type="foo" id="id:test:foo::1">
```
geirst: Please use json for new feed files.

tests/search/bugs/map-and-ngram/bug_map_and_ngram.rb
arnej27959: now using feed()

tests/search/bugs/map-and-ngram/feed.xml
arnej27959: done

arnej27959:
@geirst please take another look

PR #167, opened 2019-06-26
Test that interleaved features are regenerated when enabling bm25 feature.

toregge:
@geirst : please review


tests/search/bm25_feature/bm25_feature.rb
```diff
@@ -24,16 +24,67 @@ def test_bm25_feature
...
-  def assert_bm25_scores
-    assert_scores_for_query("content:a", [score(2, 3, idf(3)),
-                                          score(3, 7, idf(3)),
-                                          score(1, 2, idf(3))])
+  def test_enable_bm25_feature
+    set_description("Test regeneration of interleaved features when enabling bm25 feature")
+    @test_dir = selfdir + "regen/"
+    deploy_output = deploy_app(SearchApp.new.sd(use_sdfile("test.0.sd")))
+    start
+    postdeploy_wait(deploy_output)
```
geirst: This should not be needed, consider remove.

tests/search/bm25_feature/bm25_feature.rb
```diff
@@ -54,6 +105,33 @@ def assert_scores_for_query(query, exp_scores)
     end
   end
 
+  def use_sdfile(sdfile)
```
geirst: I don't think this is needed anymore. I think the sd-file can have a different name than the search / document tag inside. Please try without. If it is not possible we should have a common function in search_test.rb instead.

tests/search/bm25_feature/bm25_feature.rb
```diff
@@ -54,6 +105,33 @@ def assert_scores_for_query(query, exp_scores)
...
+    dest_sd
+  end
+
+  def postdeploy_wait(deploy_output)
+    wait_for_application(vespa.container.values.first, deploy_output)
+    wait_for_config_generation_proxy(get_generation(deploy_output))
+    wait_for_reconfig(600)
+  end
+
+  def redeploy(sdfile)
```
geirst: Is it sufficient to use redeploy() from search_test.rb instead? If not, we should extend that instead of re-implementing this several places.

tests/search/bm25_feature/bm25_feature.rb
toregge: [14:01:46.549] Invalid application package: default.default: Error loading model: Search definition file name ('test.0') and name of search element ('test') are not equal for file 'test.0.sd'

PR #178, opened 2019-07-08
Bjorncs/https

bjorncs:


lib/rest_api.rb
```diff
@@ -24,43 +24,37 @@ def http_request_delete(uri, params = {})
...
-    while iterations < max_iterations do
-      begin
-        Net::HTTP::Proxy(proxy_host, proxy_port).start(uri.host, uri.port) do |http|
-          http.open_timeout = params[:open_timeout] ? params[:open_timeout] : 4 * 60
-          http.read_timeout = params[:read_timeout] ? params[:read_timeout] : 4 * 60
-          if (params[:request])
-#            puts ":request in params #{params[:request]}"
+    @https_client.with_https_connection(original_uri.host, original_uri.port, original_uri.path, original_uri.query) do |conn, uri|
+      while iterations < max_iterations do
+        begin
```
vekterli: May retry semantics be adversely affected by us now creating the connection once outside the loop and then operating multiple times on it, rather than creating a new client per retry? I'm not 100% on how much work is actually done by the Ruby HTTP impl on the client creation itself vs. each `Net::HTTP::Get` invocation.

bjorncs: Fixed. The retry logic was most likely incorrect when the connection was created outside the retry loop. I have moved the call to `with_https_connection` inside the loop in a new commit.

PR #179, opened 2019-07-08
Garbage collect unused/non-working row-column code

hmusum:
Review only, will merge when I can observe any unwanted side effects.

@aressem FYI


hmusum:
@aressem FYI

PR #184, opened 2019-07-08
add float tensor case

havardpe:
'as float as possible' right now, which is none. query profiles ignore
cell type and .sd files do not allow it.


havardpe:
@arnej27959 please review

PR #198, opened 2019-07-10
Bjorncs/https

bjorncs:


tests/cloudconfig/orchestrator/orch_cluster.rb
```diff
@@ -32,28 +32,25 @@ def orch_uri(path)
...
+  def orch_suspend(host)
     uri = orch_uri("hosts/#{host}/suspended")
-    req = func.new uri.request_uri
-    Net::HTTP.start(uri.host, uri.port) do |session|
-      response = session.request req
-      puts "HTTP response code: #{response.code}"
-      puts "HTTP response body: \n>>>>> #{response.body} <<<<<"
-      return response
-    end
+    process_response(https_client.put(uri.host, uri.port, uri.path, nil))
```
vekterli: Could probably skip explicit `nil` for `headers` argument since it's defaulted (unless `nil` vs `{}` has special meaning?)

bjorncs: `nil` is for the body parameter that does not have a default value.

vekterli: My mistake, got my function signatures mixed up 🙂

PR #259, opened 2019-08-08
New system test testing xgboost integration and scoring with toy data…

jobergum:
… sets from sklearn


tests/search/xgboost/xgboost.rb
```diff
@@ -0,0 +1,70 @@
...
+
+class XGBoostServing < IndexedSearchTest
+
+  def setup
+    @valgrid = false
+    set_owner("jobergum")
+    set_description("Test XGBoost model representation in Vespa")
+
+    @node = @vespa.nodeproxies.values.first
+    @node.execute("pip3 install xgboost sklearn")
```
aressem: It would be best to have these as part of our systemtest image instead of installing stuff in the test itself. But, I guess it's ok for now.

jobergum: I chose this approach to test with whatever is the newest stable release in case of changes.

PR #265, opened 2019-08-13
Simplify test

hmusum:
Now just tests that header and body in search definition is allowed
and app works.


baldersheim:
or @vekterli PR

PR #276, opened 2019-08-19
Avoid writing into the system test install tree and use a temporary d…

aressem:
…irectory when adding files to the app package.


jobergum:
Great! Thanks a lot!

PR #279, opened 2019-08-23
Update expected md5

baldersheim:
@hmusum or @vekterli PR


baldersheim:
Confirmed with @geirst That this was OK

PR #304, opened 2019-09-09
Add performance tests for merge with varying offsets.

baldersheim:
@bjorncs or @bratseth PR


tests/performance/dispatch/dispatch.rb
```diff
@@ -0,0 +1,80 @@
...
+      {
+        :x => 'legend',
+        :y => 'memory.rss',
+        :historic => true
+      }
+    ]
+    num_docs = 100000
+    deploy_app(get_app())
+    container = (vespa.qrserver["0"] or vespa.container.values.first)
+    container.execute("g++ -Wl,-rpath,#{Environment.instance.vespa_home}/lib64/ -g -O3 -o #{dirs.tmpdir}/docs #{selfdir}/docs.cpp")
```
bjorncs: Consider generating documents in Ruby instead. Any specific reason for using a C program for that purpose?

tests/performance/dispatch/dispatch.rb
```diff
@@ -0,0 +1,80 @@
...
+    start
+    container.execute("#{dirs.tmpdir}/docs #{num_docs} | vespa-feeder")
```
bjorncs: Consider using the Ruby `feed()` method instead.

PR #310, opened 2019-09-09
Use ranking instead of sorting to compare performance as that is more…

baldersheim:
… common.

@bratseth or @bjorncs or @hmusum PR


tests/performance/dispatch/query.txt
```diff
@@ -1 +1 @@
-/search/?query=sddocname:test&sortspec=score
+/search/?query=sddocname:test
```
hmusum: No "&ranking=score" here?

baldersheim: Good point.

PR #324, opened 2019-09-16
Delete old unused code.

aressem:
This is the only place where we have found SIGINT to be trapped. Try to delete this file in case some of the code somehow get's it included and prevent killing processes with SIGINT.


aressem:
@tegge or @baldersheim please review and merge so we can progress with the investigation on the hanging performance test.

baldersheim:
Sorry, in one ear, out the other.

PR #336, opened 2019-09-19
Revert "Revert "Update tests so that it knows it will not go through fdispatch.""

baldersheim:
Reverts vespa-engine/system-test#334
@bratseth PR


bratseth:
👍

PR #372, opened 2019-10-11
Expect default NaN

bratseth:
@lesters this is the same change as you did in the unit test so I assume you have verified it. Merging to fix the build.


lesters:
Thanks :+1:  !

PR #373, opened 2019-10-12
Require https with redirects

hakonhall:


hakonhall:
Not sure whether we should REQUIRE https, or accept both http and https for this system test. @bjorncs?

bjorncs:
@hakonhall: Only handling https is fine.

PR #385, opened 2019-10-25
Gjoranv/log searcher info

gjoranv:


tests/container/classloading_in_deconstruct/src/main/java/com/yahoo/vespatest/DeconstructSearcher.java
```diff
@@ -35,7 +35,8 @@ public void deconstruct() {
         try {
             Class<?> causesException = this.getClass().forName("com.yahoo.docproc.DocumentProcessor");
         } catch (ClassNotFoundException e) {
-            log.log(Level.SEVERE, "Class not found when deconstructing searcher", e);
+            log.log(Level.SEVERE, "Class not found when deconstructing searcher " + this
```
hmusum: I thought the keyword was `self, not `this`

gjoranv: This is java :)

PR #388, opened 2019-10-28
Use weightedset item ordering independent results set comparisons

vekterli:
`Resultset` ends up normalizing based on key, so works transparently.


vekterli:
@arnej27959 please review (forgot to ping in original comment)

PR #404, opened 2019-11-05
Update test after maxWeight feature was added

hmusum:
Update after changes in https://github.com/vespa-engine/vespa/pull/11212
Please check if the minor changes to feature values are OK


geirst:
The feature values have not really changed, it is minor changes in double representation. This is handled in the system test via `def assert_features(expected, actual, eps = 1e-6)`.

PR #410, opened 2019-11-11
Revert "Test feature access"

hmusum:
Reverts vespa-engine/system-test#409

Two tests are now called test_tensor_in_summaryfeatures, looks like only the new one is run (and it is failing)


jonmv:
`ruby`

PR #435, opened 2019-12-03
Improve matched elements only tests

geirst:
@toregge please review


geirst:
Depends on fix in https://github.com/vespa-engine/vespa/pull/11484.

PR #442, opened 2019-12-05
Test that query tensor can have different cell type than attribute te…

geirst:
…nsor (float vs double).

@arnej27959 please review


geirst:
Depends on https://github.com/vespa-engine/vespa/pull/11514.

PR #512, opened 2020-01-21
Lesters/add matrix matrix product test

lesters:
@havardpe Please review.

For some reason, on my machine the `512_float` and `512_float_non_inner` have equal latencies. However, when I did the same with doubles (not included here), the `512_double` was more than 2x of `512_double_non_inner`. I can't see why given the code below. Bug?


havardpe:
I do not know. The code should be templated on float/double and be the same in both cases...

PR #542, opened 2020-02-03
Update config def checksum after imported fields have been added

vekterli:
@geirst please review

Change just adds a field to the config, which is safe.


geirst:
👍

PR #553, opened 2020-02-13
Use another constructor

hmusum:
Follow changes in https://github.com/vespa-engine/vespa/pull/12161


hmusum:
@baldersheim please take a look

PR #610, opened 2020-03-18
Revert "Bratseth/search to schema take 2" MERGEOK

bratseth:
Reverts vespa-engine/system-test#608

Some performance tests too


hmusum:
👍

PR #618, opened 2020-03-19
Revert "Print search chains with current syntax (<chain>)."

hmusum:
Reverts vespa-engine/system-test#615

Breaks tests:
`REXML::ParseException: Missing end tag for 'Chain:0x00000000a8f9f8' (got "search")`


geirst:
Was already fixed in https://github.com/vespa-engine/system-test/pull/617

PR #628, opened 2020-03-25
Use two nodes for the content cluster

bratseth:
Content clusters with just one node are not taken offline
any more, so use two nodes.


bratseth:
Or @hmusum

PR #629, opened 2020-03-25
Revert "Use two nodes for the content cluster"

bratseth:
Reverts vespa-engine/system-test#628

Test should work now.


bratseth:
Wrong pr ...

PR #646, opened 2020-03-31
Fix feed_params

hmusum:
Declare feed_params correctly after changes in https://github.com/vespa-engine/system-test/pull/645


baldersheim:
👍

PR #679, opened 2020-04-24
Enable distributor B-tree database for system tests (take 3)

vekterli:
@baldersheim please review

Override will be removed once feature has been made default.

Now with more database optimizations™!


baldersheim:
Merge at will.

PR #691, opened 2020-04-29
Revert "Add support for generic service tags in app generator"

aressem:
Reverts vespa-engine/system-test#687

We already have a GenericServices class in our tests.


vekterli:
The weighted set feeding perf test is going to have to be release-annotated since it depends on this functionality.

PR #692, opened 2020-04-29
Revert "Add container warmup phases using null feed endpoint service"

aressem:
Reverts vespa-engine/system-test#688


aressem:
Already reverted the change in the framework, so revert this one to to avoid failing tests.

PR #732, opened 2020-05-19
Revert "Log to see if coverage degradation is the problem."

baldersheim:
Reverts vespa-engine/system-test#731
Not necessary


hmusum:
👍

PR #745, opened 2020-05-26
Revert "Set agentpath in config server jvm args for performance tests"

aressem:
Reverts vespa-engine/system-test#698

This is not compatible with the open source system tests. We have no libperfmap.so there.

FYI, @bjorncs , @baldersheim

### DIFF
--- a/lib/environment_base.rb
+++ b/lib/environment_base.rb
@@ -74,7 +74,6 @@ class EnvironmentBase
   end
 
   def start_configserver(testcase)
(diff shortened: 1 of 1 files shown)


hmusum:
Is there a flag or something that can be used to determine if we are running internal system tests? Or perhaps this should be moved to lib-internal/environment.rb in internal repo ...

aressem:
Override this in lib-internal/environment.rb is the correct solution.

bjorncs:
Thanks! Webtide need this to run some of our performance tests. We are profiling a Jetty performance regression.

PR #770, opened 2020-06-05
Remove garbage collection performance test.

toregge:
@geirst : please review


toregge:
Test was removed due to garbage collection being throttled and managed by distributor since vespa 5.1. Thus is did no longer measure garbage collection performance, but instead the garbage collection pace set by the distributor.

Additionally, network packet loss between system tests controller (where system test script runs) and system test node (where vespa runs) could cause measured garbage collection pace to vary wildly.

PR #782, opened 2020-06-12
 Recenter after CMS

hmusum:


hmusum:
Fixed in https://github.com/vespa-engine/system-test/pull/781

PR #797, opened 2020-06-24
A name string is not a metric, but a parameter

jonmv:
@bratseth please review and merge.


jonmv:
I don't think it's used for anything, really.

jonmv:
.

PR #803, opened 2020-07-01
Increase barrier timeout, we get timeouts occasionally

hmusum:


hmusum:
Looks like Travis is having issues, merging.

PR #806, opened 2020-07-02
Enable content node B-tree bucket DB for tests

vekterli:
@baldersheim please review

Use explicit config override since there is no feature flag for
this yet (and overrides also apply when tests are run locally on
a dev container).

Also remove distributor B-tree config since it's enabled by
default now.


vekterli:
Will observe tests and revert if there are any regressions

PR #849, opened 2020-08-03
Allow higher throughput

hmusum:


hakonhall:
Recentering would mean you change the lower limits too?

hmusum:
Yes, sorry about misleading commit message, I originally thought I should change lower limit for one case, but looking longer back in history I saw that would not be right.

PR #885, opened 2020-09-11
Use 10 distribution bits for performance tests.

baldersheim:
@vekterli PR

### DIFF
--- a/lib/performance_test.rb
+++ b/lib/performance_test.rb
@@ -57,9 +57,8 @@ class PerformanceTest < TestCase
   end
 
   def deploy_app(app, deploy_params = {})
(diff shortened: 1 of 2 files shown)


lib/performance_test.rb
```diff
@@ -56,6 +56,21 @@ def get_script_user
     user
   end
 
+  def deploy_app(app, deploy_params = {})
+    # Override distribution bits to ensure no whole-corpus streaming
+    # searches have to visit 64k buckets, but only 2.
```
vekterli, line 61: Consider removing this comment, since it only makes sense with 1 distribution bit

vekterli (approved):
:+1:

PR #894, opened 2020-09-15
Remove sleep statements

hmusum:
If waiting is necessary we should do it some other way. Test works for me without sleeps


geirst (approved):
👍

PR #925, opened 2020-09-25
LibreSSL does not otuput X509 extension section on OS X. Skip testing…

aressem:
… for it there.


vekterli:
I think the underlying issue here might be that OpenSSL seems to implicitly set a `CA:TRUE` extension when generating a self-signed certificate, while LibreSSL does not. I'll try adding explicit extensions to the root CA certificate to see if that fixes things.

vekterli:
Looks like OpenSSL makes doing this the Right Way(tm) a massive pain when using command line tools, so let's go with this pragmatic solution for now 😬 Will keep a mental note to make things better.

vekterli (approved):
:+1:

PR #932, opened 2020-09-27
Stop using obsolete URI.escape for the whole URI. Use CGI.escape inst…

aressem:
…ead for the query part.

This avoids the excessive obsoletion console output in ruby 2.7.


hmusum (approved):

aressem:
Need to investigate the failing test

PR #962, opened 2020-10-05
visit at .../ns/type/docid instead of ../ns/type/number which was acc…

jonmv:
…identally supported

@vekterli please review.

### DIFF
--- a/tests/vds/documentapi/document_api_v1_part1.rb
+++ b/tests/vds/documentapi/document_api_v1_part1.rb
@@ -10,10 +10,10 @@ class DocumentApiVdsPart1 < DocumentApiV1Base
   end
 
   def test_visit_with_number
-    response = api_http_get('/document/v1/storage_test/music/number/')
+    response = api_http_get('/document/v1/storage_test/music/docid/')
     # Visit completes before covering the entire global bucket space, hence the continuation token.
     assert_json_string_equal(
-      '{"documents":[{"id":"id:storage_test:music:n=1:8", "fields":{"title":"title"}}],"pathId":"/document/v1/storage_test/music/number/",' +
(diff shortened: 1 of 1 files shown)


tests/vds/documentapi/document_api_v1_part1.rb
```diff
@@ -10,10 +10,10 @@ def test_get_with_number
   end
 
   def test_visit_with_number
-    response = api_http_get('/document/v1/storage_test/music/number/')
+    response = api_http_get('/document/v1/storage_test/music/docid/')
```
vekterli, line 13: Does test name `test_visit_with_number` make sense anymore? Seems more like  `test_continuation_token_returned_after_visiting_single_document`

vekterli (approved):
:+1: minor naming question

tests/vds/documentapi/document_api_v1_part1.rb
jonmv: Well, the document is stored with a number in its ID, so I thought leaving it like this was OK. You're the expert here, though :)

vekterli: It's not testing `/number/` anymore (and arguably never really did), so in that case it should be renamed `test_visit_of_document_with_number_location_in_id` or similar to avoid ambiguities. If we have another test that explicitly tests using `/number/` I'd still suggest renaming to the first suggestion.

jonmv: Perhaps this test was meant to include the number ...

jonmv: Ah, well it does that a few lines further down. I guess _that_'s reason to leave the name in place :)

PR #963, opened 2020-10-05
Status code 412 indicates condition not met — ne need for regex

jonmv:
@vekterli please reivew.


vekterli (approved):
:+1: out of curiosity, what are we returning now as part of the response body?

jonmv:
`"pathId": ...` and `"message": <thing from backend>`, like:
`[UNKNOWN(251013) @ tcp/vespa5.factory.corp.gq1.yahoo.com:19112/default]: ReturnCode(TEST_AND_SET_CONDITION_FAILED, Document does not exist partition=0, nodeIndex=0 bucket=20bcd79478b1e483 ) `

jonmv:
Could prepend a more humanised error message if you think that's valuable.

vekterli:
Looks good the way it is 🦄

jonmv:
Unicorn

PR #990, opened 2020-10-11
Allow override of VESPA_HOME and system test location.

toregge:
If vespa version ends with 0 then it is a snapshot version.

### DIFF
--- a/bin/run-tests-on-swarm.sh
+++ b/bin/run-tests-on-swarm.sh
@@ -179,7 +179,13 @@ readonly BASEDIR=/tmp/testresults
 readonly NETWORK="$USER-vespa"
 readonly TESTRUNNER="$USER-testrunner"
 readonly SERVICE="$USER-vespanode"
-readonly VESPAVERSION=$(docker run --rm $BINDMOUNT_OPTS --entrypoint env $DOCKERIMAGE /opt/vespa/bin/vespa-print-default version)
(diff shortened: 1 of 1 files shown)


bin/run-tests-on-swarm.sh
```diff
@@ -179,7 +179,13 @@ readonly BASEDIR=/tmp/testresults
 readonly NETWORK="$USER-vespa"
 readonly TESTRUNNER="$USER-testrunner"
 readonly SERVICE="$USER-vespanode"
-readonly VESPAVERSION=$(docker run --rm $BINDMOUNT_OPTS --entrypoint env $DOCKERIMAGE /opt/vespa/bin/vespa-print-default version)
+VESPAVERSION=$(docker run --rm $BINDMOUNT_OPTS --entrypoint bash $DOCKERIMAGE -lc '${VESPA_HOME-/opt/vespa}/bin/vespa-print-default version')
```
aressem, line 182: Need to use double quotes to get $VESPA_HOME evaluated in this scripts context.

toregge: The VESPA_HOME environment variable is supposed to be picked up from the image and be expanded by the bash process running in the container.

aressem: Right.

aressem (approved):

PR #1007, opened 2020-10-13
Add performance test for mixed tensor operations.

geirst:
@havardpe or @arnej27959 please review

### DIFF
--- a/tests/performance/mixed_tensors/mixed_tensors.rb
+++ b/tests/performance/mixed_tensors/mixed_tensors.rb
@@ -1,0 +1,84 @@
+# Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+require 'performance_test'
+require 'app_generator/search_app'
+require 'performance/fbench'
+require 'pp'
+require 'environment'
+
+class MixedTensorPerfTest < PerformanceTest
+
+  FBENCH_RUNTIME = 30
+  LABEL = "label"
+  MIXED_BASIC = "mixed_basic"
+
+  def initialize(*args)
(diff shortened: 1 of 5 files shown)


tests/performance/mixed_tensors/mixed_tensors.rb
```diff
@@ -0,0 +1,84 @@
...
+
+  def compile_data_gen
+    @data_gen = dirs.tmpdir + "data_gen"
+    @container.execute("g++ -Wl,-rpath,#{Environment.instance.vespa_home}/lib64/ -g -O3 -o #{@data_gen} #{selfdir}/data_gen.cpp")
+  end
+
+  def gen_and_copy_query_file(num_queries)
+    @query_file = dirs.tmpdir + "queries.txt"
+    @container.execute("#{@data_gen} queries #{num_queries} > #{@query_file}")
+    @container.copy(@query_file, File.dirname(@query_file))
```
arnej27959, line 48: this looks un-needed and wrong, since you generated the file on the container node, but "copy" assumes it was generated locally and should be copied to the container node.

arnej27959 (commented):
i'm never sure before it actually works in RL but i think i spotted a bug...

tests/performance/mixed_tensors/mixed_tensors.rb
geirst: Yes, I think you are right. Nice catch. Let's try to remove this line.

geirst: Fixed.

arnej27959 (approved):
LGTM

PR #1012, opened 2020-10-14
Fix expectations in test according to updated /document/v1 handler

jonmv:
@bjorncs please review.


bjorncs (approved):

jonmv:
Think we'll just annotate the test if we have to revert the handler. No work currently in progress on the old handler ...

PR #1017, opened 2020-10-14
Add explicit multi bucket space node restart test

vekterli:
@geirst please review

We have lots of tests implicitly testing data retention upon node
restart, but want to have an _explicit_ test that also spans both the
default and the global bucket space.


geirst (approved):
👍

PR #1027, opened 2020-10-15
Add simple performance test for document v1

jonmv:
@bjorncs please review briefly. Draft for now — I would like to verify locally as well :)

### DIFF
--- a/tests/performance/document_rest_api/document_v1_throughput.rb
+++ b/tests/performance/document_rest_api/document_v1_throughput.rb
@@ -1,0 +1,91 @@
+# Copyright 2020 Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
+require 'performance_test'
+require 'app_generator/search_app'
+
+class DocumentV1Throughput < PerformanceTest
+
+  def initialize(*args)
+    super(*args)
+  end
+
+  def timeout_seconds
+    900
+  end
+
+  def setup
+    super
+    set_description("Stress test document/v1 API POST and GET")
+    set_owner("jvenstad")
+  end
+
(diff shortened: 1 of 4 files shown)


jonmv:
Works and ready for review and merge. PTAL @bjorncs or @geirst  :)

jonmv:
Hmm, this is weird, though: all documents should be present, but this is the breakdown for the GET part. POST part has all 200 OKs and more than 65k successes ... 
```
[14:25:51.311]        200 :   326144
[14:25:51.311]        404 :    64458
```

jonmv:
Aha, looks like an `fbench` artefact: the ratio is even worse for the old handler, which manages only half as many PUTs — apparently, the fbench clients aren't cooperatively looping through the data multiple times.

tests/performance/document_rest_api/document_v1_throughput.rb
```diff
@@ -0,0 +1,95 @@
...
+
+    start
+    benchmark_operations(@test_config)
+  end
+
+  def benchmark_operations(methods)
+    qrserver = @vespa.container["combinedcontainer/0"]
+    methods.each do |method, config|
+      paths_file = dirs.tmpdir + method + ".txt"
+      qrserver.execute("for i in {1..#{1 << 16}};"\
```
bjorncs: Consider writing converting this inline shell script to Ruby.

bjorncs (approved):

tests/performance/document_rest_api/document_v1_throughput.rb
jonmv: Yes, will consider. This is still WIP :)

PR #1029, opened 2020-10-15
Don't check for disk in bucketdb dump, as it no longer exists

vekterli:
@baldersheim please review this change in our highly sophisticated™ bucket DB output matching


geirst (approved):
👍

baldersheim (approved):

baldersheim:
Thanks, I had just started looking at it. It would have taken long....

PR #1034, opened 2020-10-16
Remove sigbushandler test.

toregge:
@baldersheim : please review


toregge:
@aressem : FYI

aressem:
I did a fix on this test in https://github.com/vespa-engine/system-test/pull/1036, but if this does not have any value anymore can stay deleted.

PR #1053, opened 2020-10-22
Skip parts of test if CentOS (missing linguistics support for CJK lan…

hmusum:
…guages)

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -7,7 +7,7 @@ RUN yum -y install --setopt=skip_missing_names_on_install=False \
                    rh-ruby25-rubygem-net-telnet devtoolset-9-gcc-c++ git && \
     source /opt/rh/rh-ruby25/enable && \
     source /opt/rh/devtoolset-9/enable && \
-    gem install libxml-ruby gnuplot distribution test-unit builder concurrent-ruby facter && \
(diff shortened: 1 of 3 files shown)


docker/Dockerfile
```diff
@@ -7,7 +7,7 @@ RUN yum -y install --setopt=skip_missing_names_on_install=False \
                    rh-ruby25-rubygem-net-telnet devtoolset-9-gcc-c++ git && \
     source /opt/rh/rh-ruby25/enable && \
     source /opt/rh/devtoolset-9/enable && \
-    gem install libxml-ruby gnuplot distribution test-unit builder concurrent-ruby && \
+    gem install libxml-ruby gnuplot distribution test-unit builder concurrent-ruby facter && \
```
aressem, line 10: This also needs to be put into the dev image:
https://github.com/vespa-engine/docker-image-dev/blob/master/dev/centos7/Dockerfile

Alternatively read the file /etc/redhat-release and look for CentOS

aressem (approved):

docker/Dockerfile
hmusum: Changed to read `/etc/redhat-release`

aressem (approved):

PR #1066, opened 2020-10-24
Allow higher latency

hmusum:
Note: Latency has increased quite a bit, so this is more than just an minor adjustment of limits.


geirst (approved):
👍

geirst:
@arnej27959 and @havardpe FYI

geirst:
The change in latency happened between 7.299.110 and 7.299.112, and the reason is this fix: https://github.com/vespa-engine/system-test/pull/1015.

PR #1068, opened 2020-10-25
run micro-benchmark for tensor instructions

arnej27959:

@havardpe or @geirst please review

### DIFF
--- a/tests/performance/tensor_eval/tensor_instructions.rb
+++ b/tests/performance/tensor_eval/tensor_instructions.rb
@@ -18,7 +18,7 @@ class TensorInstructionBenchmarkTest < PerformanceTest
 
   def test_instruction_benchmark
     set_description("Test performance of low-level tensor instructions")
-    deploy_app(SearchApp.new.sd(selfdir+"foobar.sd"))
+    deploy_app(SearchApp.new.sd(selfdir+"dummy.sd"))
     searchnode = vespa.search['search'].first
     output = searchnode.execute("#{Environment.instance.vespa_home}/bin/vespa-tensor-instructions-benchmark")
     @graphs = []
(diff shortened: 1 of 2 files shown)


havardpe (approved):
I am a little worried that this will break the graph system

tests/performance/tensor_eval/tensor_instructions.rb
```diff
@@ -0,0 +1,87 @@
...
+  BM_TYPE = 'benchmark_type'
+
+  def setup
+    super
+    set_owner("arnej")
+  end
+
+  def test_instruction_benchmark
+    set_description("Test performance of low-level tensor instructions")
+    deploy_app(SearchApp.new.sd(selfdir+"foobar.sd"))
```
geirst, line 21: Consider renaming to `test.sd` and add the file to the PR.

tests/performance/tensor_eval/tensor_instructions.rb
```diff
@@ -0,0 +1,87 @@
...
+    searchnode = vespa.search['search'].first
+    output = searchnode.execute("#{Environment.instance.vespa_home}/bin/vespa-tensor-instructions-benchmark")
```
geirst, line 23: I think we need an option to the program where we run with fewer implementations and fewer benchmark cases. Currently we have ~80 benchmark cases and 6 implementations. Could also consider to graph only a subset.

tests/performance/tensor_eval/tensor_instructions.rb
arnej27959: there's already a "test.sd" there, but havardpe suggested "dummy.sd" which makes it clear it's just a placeholder.

tests/performance/tensor_eval/tensor_instructions.rb
arnej27959: yes, that's WIP in another PR...

tests/performance/tensor_eval/tensor_instructions.rb
geirst: 👍

tests/performance/tensor_eval/tensor_instructions.rb
geirst: 👍

PR #1091, opened 2020-11-04
Use minimum of 8 distribution bits in tests and remove use of legacy distribution type 

vekterli:
@geirst please review


vekterli:
Forgot to update expected app generator test outputs, will do that tomorrow

geirst (approved):
👍

PR #1092, opened 2020-11-04
Maybe filter key needs to be quoted?

hmusum:


jonmv:
My bad, 'runtime' is already filled in — it was only the `legend` x axis that was problematic.

jonmv:
Now you don't get data, because the runtime key is dupliacted acress metrics and parameters ...

PR #1102, opened 2020-11-06
Verify reindexing is run after index mode change

jonmv:
@bjorncs and @geirst please review.

### DIFF
--- a/tests/search/schemachanges/schemachanges_need_refeed_reconfig.rb
+++ b/tests/search/schemachanges/schemachanges_need_refeed_reconfig.rb
@@ -41,7 +41,7 @@ class SchemaChangesNeedRefeedReconfigTest < IndexedSearchTest
     # Wait up to 2 minutes for reindexing to be ready
     tenant = use_shared_configservers ? @tenant_name : "default"
     application = use_shared_configservers ? @application_name : "default"
-    reindexingUrl = "https://#{vespa.configservers["0"].name}:#{vespa.configservers["0"].ports[1]}/application/v2/tenant/#{@tenant_name}/application/#{@application_name}/environment/prod/region/default/instance/default/reindexing"
+    reindexingUrl = "https://#{vespa.configservers["0"].name}:#{vespa.configservers["0"].ports[1]}/application/v2/tenant/#{tenant}/application/#{application}/environment/prod/region/default/instance/default/reindexing"
     startTime = Time.now
     until Time.now - startTime > 120 # seconds
       response = http_request(URI(reindexingUrl), {})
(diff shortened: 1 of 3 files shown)


jonmv:
Hmm, probably also need to wait explocitly for config convergence, and then another minute dor maintainer to pick this up, before doing redeployment.

@bjorncs should we also let the maintainer trigger redeployment when reindexing is ready ...?

tests/search/schemachanges/schemachanges_need_refeed_reconfig.rb
```diff
@@ -43,6 +43,14 @@ def test_need_refeed_reconfig
     assert_result("sddocname:test&nocache", @test_dir + "result.1.xml")
     assert_hitcount("f1:b&nocache", 2)
     assert_hitcount("f3:%3E29&nocache", 2)
+
+    # redeploy again to trigger reindexing, then wait for up to 5 minutes for document 1 to be reindexed
+    redeploy("test.1.sd")
+    start = Time.now
+    until /<field name="a1">30<\/field>/ =~ search("sddocname:test&nocache") or Time.now - start > 300 # seconds
```
geirst, line 50: I don't think we should rely on regexp matching here. It is better to use the Resultset object and find the hit and field value there.

bjorncs (approved):

tests/search/schemachanges/schemachanges_need_refeed_reconfig.rb
jonmv: Thanks.

jonmv:
Should also wait for container nodes, which run the indexing script, to have the new config generation. This done already, in `redeploy`, for content nodes, but not for containers.

jonmv:
The config server itself wait for config convergence of the whole app, before setting reindexing as ready. Checking this API will suffice. 

Also found the `Resultset`, @geirst 👍

PTAL.

tests/search/schemachanges/schemachanges_need_refeed_reconfig.rb
```diff
@@ -39,7 +39,8 @@ def test_need_refeed_reconfig
     assert_match(/Consider re-indexing document type 'test' in cluster 'search'.*\n.*Field 'f2' changed: add index aspect/, redeploy_output)
 
     # Wait up to 2 minutes for reindexing to be ready
-    until Time.now - start > 120 # seconds
+    startTime = Time.now
```
geirst, line 45: Please use snake case for ruby code: start_time.

tests/search/schemachanges/schemachanges_need_refeed_reconfig.rb
```diff
@@ -39,9 +39,10 @@ def test_need_refeed_reconfig
     assert_match(/Consider re-indexing document type 'test' in cluster 'search'.*\n.*Field 'f2' changed: add index aspect/, redeploy_output)
 
     # Wait up to 2 minutes for reindexing to be ready
+    reindexingUrl = "https://#{vespa.configservers["0"].name}:#{vespa.configservers["0"].ports[1]}/application/v2/tenant/#{@tenant_name}/application/#{@application_name}/environment/prod/region/default/instance/default/reindexing"
```
geirst, line 44: Please use snake case for ruby code: reindexing_url

jonmv: Of course!

jonmv:
Can't get the config server API to work. Need to figure out this.

jonmv:
PTAL.

jonmv:
PTAL.

PR #1103, opened 2020-11-06
Improved performance after storing tensors directly in-memory.

geirst:
@arnej27959 or @aressem or @hmusum or @baldersheim please review and merge.


arnej27959 (approved):
👍

PR #1109, opened 2020-11-11
Change tensor type of tensor attribute no longer requires refeed MERGEOK

bjorncs:


hmusum (approved):

arnej27959 (changes requested):
i need to talk to Geir but I don't see any way this can work...

arnej27959:
we need to at least keep the test, if we fix things so DocumentTypeChangeValidator can catch this change we may need to update it to match a new expected message.

PR #1121, opened 2020-11-15
Fix int vs string assertion

lesters:
@arnej27959 Please review. When conformance passes, exit code is currently "0".


arnej27959 (approved):
right, slightly unexpected that...

PR #1128, opened 2020-11-18
Jonmv/reindexing test for indexing mode change

jonmv:
@geirst please review. Only the last commit is new from #1102


geirst (approved):
👍

PR #1130, opened 2020-11-19
Add perf test case where multiple models are put into the same mixed …

geirst:
…tensor.

@arnej27959 please review
@havardpe FYI

### DIFF
--- a/tests/performance/mixed_tensors/data_gen.cpp
+++ b/tests/performance/mixed_tensors/data_gen.cpp
@@ -103,30 +103,32 @@ void print_puts(std::ostream& os, const StringVector& models, RandomStrings& str
 const std::string LB = "%7B";
 const std::string RB = "%7D";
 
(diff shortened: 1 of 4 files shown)


arnej27959 (changes requested):
I don't think this is the most useful thing for us to benchmark.
In the multi_model your tensors are now all type:  `tensor(model{},cat{})`
which means that it's just the same computation as single_model, but with more
data in the first input, and a slight extra cost in each step because the sparse address is two `vespalib::stringref`s instead of one.

I would suggest keeping most of the computation equal to the single_model but add
an extra step

```
# result: tensor(cat{},vec[256]),
macro inline select_model_version() {
  expression: reduce(query(q_select_model) * attribute(models), sum, model)
}
```
where `q_select_model` is `tensor(model{})` and in the query generation you just select one random among the model (and cell value 1.0).

does that make sense to you?

geirst:
Yes, this is a good suggestion. PTAL.

arnej27959 (approved):
looks good, looking forward to perf reports from this

PR #1142, opened 2020-11-23
Use two rank-profiles for the multi-model case, one with early reduce…

geirst:
… and one with only late reduce.

@arnej27959 please review
@havardpe FYI


arnej27959 (approved):
could be interesting 👍

PR #1150, opened 2020-11-25
Write to feeder node, and address review comments

jonmv:
@geirst please review.

### DIFF
--- a/tests/performance/reindexing/reindexing_and_feeding.rb
+++ b/tests/performance/reindexing/reindexing_and_feeding.rb
@@ -5,150 +5,112 @@
 require 'performance_test'
 require 'app_generator/search_app'
 
-class ReindexingAndFeeding < PerformanceTest
+class ReindexingAndFeedingTest < PerformanceTest
 
   def initialize(*args)
     super(*args)
   end
 
   def timeout_seconds
-    900
+    1800
   end
 
   def setup
-	  super
-	  set_description("Measure throughput of reindexing, and its impact on external updates and puts")
-	  set_owner("jvenstad")
-	  @test_config = {
-		  {
-			  :legend => 'Reindexing',
-			  :fbench => { :clients => 4, :use_post => true },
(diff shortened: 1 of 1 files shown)


jonmv:
Ah, ruby devz ...

tests/performance/reindexing/reindexing_and_feeding.rb
```diff
@@ -5,150 +5,112 @@
...
+		gateway(ContainerDocumentApi.new)).
+    admin_metrics(Metrics.new).
+    indexing("combinedcontainer").
+    sd(selfdir + "doc.sd")
+
     deploy_app(@app)
     start
 
+    @qrserver = @vespa.container["combinedcontainer/0"]
+    @document_count = 1 << 19
```
geirst, line 41: I prefer an explicit number here.

tests/performance/reindexing/reindexing_and_feeding.rb
```diff
@@ -193,21 +155,75 @@ def get_reindexing_status
...
   end
 
+  def graph_config
+    {
+      'reindex' => {
+	'reindexing.throughput'   => { :y_min =>  10000, :y_max =>  50000 },
+	'reindexing.time.seconds' => { :y_min =>     30, :y_max =>    120 }
+      },
+      'feed' => {
+	'qps' =>                     { :y_min =>  10000, :y_max =>  50000 },
```
geirst, line 165: Is it feed throughput you want here? If so, _feeder.throughput_ is the key used when generating the report.

jonmv: Thanks, yes. `qps` is used by fbench :)

tests/performance/reindexing/reindexing_and_feeding.rb
jonmv: I like powers of two, and so does bucket distribution, but OK :)

PR #1158, opened 2020-11-26
Test that document with predicate field can be re-fed correctly

bjorncs:

### DIFF
--- a/tests/search/booleansearch/booleansearch.rb
+++ b/tests/search/booleansearch/booleansearch.rb
@@ -417,14 +417,14 @@ class BooleanSearchTest < SearchTest
   def test_predicate_optimizations_are_idempotent
     set_description("Test that predicate optimizations are idempotent")
 
-    unoptimized_predicate = "not (not (gender in ['Male'] and age in [20..29] and true)) and gender not in ['Female']"
+    unoptimized_predicate = "not (not (gender in ['Male'] and age in [20..29] and true)) and country not in ['Sweden']"
     doc_id = "unoptimized-1"
     File.open(@feed_file, "w") {|file| write_doc(file, doc_id, unoptimized_predicate) }
     deploy_and_feed
 
     container_port = Environment.instance.vespa_web_service_port
     optimized_doc = vespa.document_api_v1.get("id:test:test::#{doc_id}", :port => container_port)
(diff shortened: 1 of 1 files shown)


tests/search/booleansearch/booleansearch.rb
```diff
@@ -410,6 +414,25 @@ def assert_search_and_document_summary_for_partial_update_test
...
+    set_description("Test that predicate optimizations are idempotent")
+
+    unoptimized_predicate = "not (not (gender in ['Male'] and age in [20..29] and true)) and gender not in ['Female']"
+    doc_id = "unoptimized-1"
+    File.open(@feed_file, "w") {|file| write_doc(file, doc_id, unoptimized_predicate) }
+    deploy_and_feed
+
+    container_port = Environment.instance.vespa_web_service_port
+    optimized_doc = vespa.document_api_v1.get("id:test:test::#{doc_id}", :port => container_port)
+    expected_optimized_predicate = "gender not in [Female] and gender in [Male] and age in [20..29]"
```
jonmv, line 427: Missing quotes for the sexes?

jonmv: I also think this could be optimised to `"gender in ['Male'] and age in [20..29]"`, but that's not necessarily what we do :)

jonmv (approved):
Approved given you consider the comments.

tests/search/booleansearch/booleansearch.rb
bjorncs: Quotes are optional and handled by the parser.  I changed the predicate now to have a clause about country instead.

jonmv:
Hehe, the comment was about the set `[Male]` being a stricter predicate than `not [Female]` ... since these are sets, not classes, right, i.e., `Male` is not a function we evaluate on a document, but rather a specific value a field should have?

jonmv:
That is, nothing which is `in [Male]` can't possibly also not be `not in [Female]`

jonmv:
Or, to be less confusing: Anything which is `in [Male]` must also be `not in [Female]`, unless `Female` and `Male` are overlapping sets .. but they aren't right?

bjorncs:
Predicate search allows a query to assign multiple values for a given key, e.g passing both `Male` and `Female` as values for `gender`. The original query made sense, but it was a bit confusing.

PR #1165, opened 2020-11-30
No loadtype metrics

baldersheim:
@vekterli PR


vekterli (approved):
:+1: does this test make sense to keep around anymore?

baldersheim:
I am not sure. I does test that you can assign loadtype as an alternative way to set priority.
But not sure if this is worth a separate test.

PR #1169, opened 2020-12-02
Print request body when non-200 status

jonmv:
@bjorncs


jonmv:
Test broke with your client.

jonmv:
That is, getting reindexing status from config server turned non-200.

bjorncs:
Ok, I'll look into it

PR #1188, opened 2020-12-08
Use realpath instead of readlink 

hakonhall:
Say /foo/slink is a symlink with content "target", and /foo/target is a file.
Then the symlink is valid and points to /foo/target.

However, if PWD is /bar and the symlink content is read, that content "target"
is NOT a valid file as there is no /bar/target file.  This is what the current
code does.  If the content of the symlink had been an absolute path name
(/foo/target), it WOULD be valid, which is why it works today.

### DIFF
--- a/lib/node_server.rb
+++ b/lib/node_server.rb
@@ -588,10 +588,10 @@ class NodeServer
     Dir.glob(expression)
   end
 
-  # Returns the name of the file referenced by the link _filename_.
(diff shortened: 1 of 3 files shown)


hakonhall:
The old code would fail if the path was not a symlink - let me change the code to preserve that feature.

hmusum (approved):

PR #1189, opened 2020-12-08
Recenter limits, lower latency

hmusum:


arnej27959 (approved):
this is Henning optimizing tensor-from-labels etc.

PR #1192, opened 2020-12-09
lower latency as expected

arnej27959:
* tensorFromLabels etc was optimized so the "slow" case is faster now.

@baldersheim or @geirst please review and merge


geirst (approved):
👍

PR #1193, opened 2020-12-09
Increase slack for max hits. Buckets can be reactivated on

toregge:
old node after having been activated on new node and
deactivated on old node.

Increase slack for max docs. Buckets that have been migrated
away from old node can be migrated back again.

@geirst : please review
@vekterli : FYI


geirst (approved):
👍

PR #1208, opened 2020-12-18
Account for clock skew on factory in reindexing test

jonmv:
@hmusum


hmusum (approved):
OK, but if we have clock skew on test nodes all sorts of strange things might happen and I'd rather fix that instead.

jonmv:
What amount of skew can we accept? This test runs a clock sensitive query immediately after feeding documents, so only a very small amount of skew will make it fail.

jonmv:
But, yes, we should have control over the clock skew. @aressem what's the status on this?

jonmv:
The test has been failing more than others — if this fixes that, it was skew.

PR #1217, opened 2021-01-05
Moved test to open source

baldersheim:
@hmusum or @aressem PR


tests/performance/feeding_and_recovery/testplot.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+lastdir=$(ls /home/y/logs/systemtests/command-line/ | sort | tail -1)
```
aressem, line 2: Probably not used ? Reference to /home/y.

aressem (changes requested):

tests/performance/feeding_and_recovery/testplot.sh
baldersheim: Good point, removed

aressem (approved):

PR #1220, opened 2021-01-05
Allow to run independently of the time of day.

baldersheim:
@hmusum or @aressem PR
It is now down to under 1800s and it seems to trigger distribution issues so it is better to run it every time.
It will probably add 5-6% to perf runtime.


hmusum (approved):
👍

PR #1221, opened 2021-01-06
Lesters/tensorflow jni removal

lesters:
@bratseth Please review. Removes all use of TensorFlow JNI from system tests.
@aressem FYI.


bratseth (approved):
👍

PR #1230, opened 2021-01-07
Test that using ngram on an external field does not affect the input …

geirst:
…field.

@bratseth please review

This depends on fix in https://github.com/vespa-engine/vespa/pull/15953.


geirst:
@bratseth ping

bratseth (approved):
👍  missed this

PR #1232, opened 2021-01-08
Recenter limits

hmusum:
Performance has changed after https://github.com/vespa-engine/vespa/pull/15951


baldersheim:
better to revert as this is unexpected. https://github.com/vespa-engine/vespa/pull/15959

PR #1244, opened 2021-01-13
Add baseline performance tests for feeding to mixed tensors.

geirst:
@arnej27959 please review
@havardpe FYI


arnej27959 (approved):
👍

havardpe (approved):
we probably need to benchmark with both number and string labels (after small numbers become direct string ids not backed by the shared store).

geirst:
Yes, I can extend the tests with that in the next iteration.

geirst:
Using both number and string labels is added here: https://github.com/vespa-engine/system-test/pull/1246

PR #1245, opened 2021-01-13
add a graph with all rank profiles

arnej27959:
* for easier comparison of latencies

@geirst please review
we can probably simplify and drop one of the new rank-profiles also.


geirst (approved):
👍

PR #1246, opened 2021-01-13
Test with both numbers and strings as labels in feed perf tests.

geirst:
@havardpe please review
@arnej27959 FYI

### DIFF
--- a/tests/performance/mixed_tensors/data_gen.cpp
+++ b/tests/performance/mixed_tensors/data_gen.cpp
@@ -5,6 +5,7 @@
 #include <iostream>
 #include <numeric>
 #include <string>
+#include <unistd.h>
 #include <vector>
 
 using StringVector = std::vector<std::string>;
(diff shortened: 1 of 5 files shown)


arnej27959 (approved):
from perf reports i think there aren't enough labels (compared to cells) so we may not see much difference anyway; if that happens we should consider changing to vectors of size 16 and have more values for the "cat" dimension.

tests/performance/mixed_tensors/data_gen.cpp
```diff
@@ -14,19 +15,23 @@ class RandomStrings {
...
     StringVector _strings;
 
-    StringVector gen_strings(size_t count, size_t offset) {
+    StringVector gen_strings(size_t count, size_t offset, bool number_string) {
         StringVector result;
         for (size_t i = offset; i < (count + offset); ++i) {
             char str[7];
-            snprintf(str, 7, "%06d", i);
+            if (number_string) {
+                snprintf(str, 7, "%06d", i);
```
geirst, line 23: I also realised that due to the leading zeroes we are not hitting the optimisation in https://github.com/vespa-engine/vespa/pull/16033. I will change that in a later PR. @arnej27959 @havardpe FYI

PR #1257, opened 2021-01-17
Adjust limits

tokle:


tests/performance/container/http/container-http.rb
```diff
@@ -95,26 +95,26 @@ def test_container_http_performance_with_logging
             :filter => {'legend' => STANDARD},
             :x => 'legend',
             :y => 'qps',
-            :y_min => 110000,
-            :y_max => 140000,
+            :y_min => 63000,
```
baldersheim, line 98: Why this drop when using persistent connections ? I thought connection logging only took place when connection was established/disconnected.

tokle: This is the baseline with access logging enabled. Connection logging is not enabled yet.

baldersheim: K

baldersheim (approved):

bjorncs (commented):
👍

PR #1258, opened 2021-01-18
Dummy first time is gone, but readyAt may be nil

jonmv:
@bjorncs please review.


jonmv:
Must be merged with https://github.com/vespa-engine/vespa/pull/16079

bjorncs (approved):

PR #1270, opened 2021-01-22
Set graph limits for mixed tensor feed tests.

geirst:
@baldersheim or @hmusum or @arnej27959 please review and merge


arnej27959 (approved):
👍

PR #1278, opened 2021-01-26
Triggering reindexing is faster, hope skew is less than 2s

jonmv:
@hmusum please review and merge.


jonmv:
(Triggering took 4.5s, so some very few documents were reindexed before the "now" timestamp >_<

hmusum (approved):

PR #1295, opened 2021-01-28
dial back maxpending

arnej27959:
* too high maxpending causes container OOM
* go back to maxpending=100, that looked stable
* tune compression level for document store

@geirst or @baldersheim please review and merge


baldersheim (approved):

geirst (commented):
👍

PR #1301, opened 2021-01-29
adjust (widen) limits

arnej27959:
* with warmup we hope that performance will be more stable,
  but use pretty wide limits in case warmup is not quite
  sufficient
* from perf reports there should no longer be much difference
  between label types, so use same limits for both type of graphs
  until we have more data
* make sure to allow the best performance we have seen so far, we
  should be able to reach that now

@geirst or @baldersheim please review and merge


geirst (approved):
👍

PR #1316, opened 2021-02-03
Revert "Use resource limits to override disk and memory"

aressem:
Reverts vespa-engine/system-test#1314

Broke system tests


toregge (approved):

geirst:
This is re-applied in https://github.com/vespa-engine/system-test/pull/1321.

PR #1325, opened 2021-02-04
If a search app has set search cluster explicitly then we cannot use

toregge:
methods that forwards to default cluster. Call methods directly on
search cluster instead.

@geirst : please review


geirst (approved):
👍

PR #1336, opened 2021-02-08
Add "host" as parameter to all performance test results

jonmv:
@aressem please review. 
@baldersheim FYI.

### DIFF
--- a/lib/performance_test.rb
+++ b/lib/performance_test.rb
@@ -216,7 +216,7 @@ class PerformanceTest < TestCase
     fillers.each do |filler|
       filler.call(@rep_result)
     end
-    @rep_result.add_parameter('host', `hostname`)
+    @rep_result.add_parameter('hosts', hostlist.join(","))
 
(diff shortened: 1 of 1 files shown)


lib/performance_test.rb
```diff
@@ -216,6 +216,7 @@ def write_report(fillers)
     fillers.each do |filler|
       filler.call(@rep_result)
     end
+    @rep_result.add_parameter('host', `hostname`)
```
aressem, line 219: This code is executed on the test runner / controller. Probably need to get the hostnames from the VespaModel and we can have more than one host.

aressem (changes requested):

lib/performance_test.rb
jonmv: Hmm ... need to change performance data query to use "matches" to support multiple values of a parameter, but can do.

jonmv: Meeeh, that may break other usages, though.

jonmv: @baldersheim what tests really use several nodes?

jonmv: Nvm, can make it work without breaking the others.

aressem (approved):

PR #1338, opened 2021-02-09
Make some warning eleminations default.

baldersheim:
@hmusum PR

### DIFF
--- a/lib/testcase.rb
+++ b/lib/testcase.rb
@@ -97,16 +97,15 @@ class TestCase
       @@log_messages[:log4j_initialize],
       @@log_messages[:log4j_see_noconfig],
       @@log_messages[:concurrent_mark_sweep],
-      @@log_messages[:zookeeper_shutdown]
+      @@log_messages[:zookeeper_shutdown],
+      @@log_messages[:async_slow_resolve],
+      @@log_messages[:slow_processing]
     ]
     @valgrind_ignorable_messages = [
-      @@log_messages[:slow_processing],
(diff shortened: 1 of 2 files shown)


lib/testcase.rb
```diff
@@ -731,7 +732,10 @@ def get_default_log_check_levels
     :log4j_initialize => /log4j:WARN Please initialize the log4j system properly/,
     :log4j_see_noconfig => /log4j:WARN See http:\/\/logging.apache.org\/log4j\/1.2\/faq.html#noconfig/,
     :concurrent_mark_sweep => /OpenJDK 64-Bit Server VM warning: Option UseConcMarkSweepGC was deprecated in version 9.0 and will likely be removed in a future release./,
-    :zookeeper_shutdown => /Starting non-reconfigurable ZooKeeper server failed on attempt/
+    :zookeeper_shutdown => /Starting non-reconfigurable ZooKeeper server failed on attempt/,
+    :time_move_backwards => /Time has moved backwards/,
```
hmusum, line 736: Did you intend to add `:time_move_backwards` to the list of ignorable messages as well?

baldersheim: No, I just wanted in the list of possible warnings. Time does move backwards from time to time with docker-on-mac.
This just makes it a bit easier ignore it when running locally.
It should not be ignored normally.

hmusum (approved):

lib/testcase.rb
hmusum: 👍

PR #1367, opened 2021-02-19
A stopped job due to too many outstanding operations will incorrectly…

baldersheim:
… seem complete.

If we require it to happen at least 3 times in a row with 1s between each sample we should probably be fine.
However a better mechanism should exist. A metric for pending buckets to move, and pending lids to compact is probably better.


toregge (approved):

vekterli (commented):
:+1:

PR #1377, opened 2021-02-22
Test that when multiple ready replicas are available, we active the one with the most documents

vekterli:
@geirst please review

This is currently not the case, so test will fail until the relevant changes
have been committed to distributor replica activation logic.


geirst (approved):
👍

PR #1388, opened 2021-03-01
Add test for bucket activation inhibition

vekterli:
@geirst please review. Have manually confirmed test fails without the appropriate config override set.

Test that bucket activation behavior can be configured to inhibit
the activation of bucket replicas whose metadata state differs from
that of the majority. Also test that all replicas are activated
as expected once they get back into sync.


geirst (approved):
👍

vekterli:
Travis failure is unrelated

PR #1392, opened 2021-03-02
Add some progress output

baldersheim:
@jonmv PR


jonmv (approved):
I guess the initial deletion was for running the test multiple times during test development, and is OK to remove.

PR #1396, opened 2021-03-03
Use fewer documents in reindexing test to avoid timeout

jonmv:
@aressem or @baldersheim or @hmusum please review and merge.


hmusum (approved):

aressem:
Great 👍

PR #1397, opened 2021-03-05
Keep a small fixed lidbloat so that documents does not move for small…

baldersheim:
… corpus sizes during tests.

@geirst PR

### DIFF
--- a/lib/app_generator/search_app.rb
+++ b/lib/app_generator/search_app.rb
@@ -13,6 +13,7 @@ class SearchApp < App
                   :garbagecollectioninterval => :garbagecollectioninterval,
                   :doc_type => :doc_type,
                   :threads_per_search => :threads_per_search,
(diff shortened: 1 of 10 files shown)


lib/app_generator/search_app.rb
```diff
@@ -13,6 +13,7 @@ class SearchApp < App
                   :garbagecollectioninterval => :garbagecollectioninterval,
                   :doc_type => :doc_type,
                   :threads_per_search => :threads_per_search,
+                  :allowedlidbloat => :allowedlidbloat,
```
geirst, line 16: Consider using snake case (as elsewhere): allowed_lid_bloat

geirst (approved):
👍

lib/app_generator/search_app.rb
baldersheim: Snake it is.

geirst (commented):
👍

PR #1409, opened 2021-03-11
Update expeced md5 after adding dictionary.ordering to attributes.def…

baldersheim:
… config

@aressem or @hmusum or @geirst or @toregge PR


hmusum:
👍

PR #1419, opened 2021-03-16
This test is timing and load sensitive, so be happy with just one val…

baldersheim:
…id non-zero sample of progress.

@geirst or @aressem or @hmusum PR


geirst (approved):
👍

PR #1420, opened 2021-03-16
Log to testoutput.log instead on the nodeserver.

baldersheim:
@vekterli PR

### DIFF
--- a/lib/nodetypes/storage.rb
+++ b/lib/nodetypes/storage.rb
@@ -156,17 +156,17 @@ class Storage
               currVersion = vdsnode.get_cluster_state_version
               max_fetch_time = [Time.now - time_started, max_fetch_time].max
               if (currVersion >= clusterstate.version)
-                #@testcase.output("Cluster state version in #{node.type}.#{index} is now #{currVersion}")
+                @testcase.output("Cluster state version in #{node.type}.#{index} is now #{currVersion}")
                 break;
               end
               if (endtime < Time.now)
                 statuspage = vdsnode.get_status_page("/systemstate")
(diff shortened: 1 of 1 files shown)


lib/nodetypes/storage.rb
```diff
@@ -156,17 +156,17 @@ def wait_for_cluster_state_propagate(clusterstate, timeout = 120, check_states =
...
               max_fetch_time = [Time.now - time_started, max_fetch_time].max
               if (currVersion >= clusterstate.version)
-                #@testcase.output("Cluster state version in #{node.type}.#{index} is now #{currVersion}")
+                @testcase.output("Cluster state version in #{node.type}.#{index} is now #{currVersion}")
                 break;
               end
               if (endtime < Time.now)
                 statuspage = vdsnode.get_status_page("/systemstate")
-                puts "Status page to get cluster state: '#{statuspage}'"
+                @testcase.output("Status page to get cluster state: '#{statuspage}'")
```
vekterli, line 164: This might be a bit much data to print for every node, as it contains the entire cluster state history the node has seen. Consider commenting out this one and keep the one emitting the actual state string.

vekterli (approved):
:+1:

lib/nodetypes/storage.rb
baldersheim: Note that this will only happen on the node failing as it throws an exception afterwards and exits.
Do you still think it is too much, or does it not contain any valuable information ?

vekterli: Ah that's true. In that case it's perfectly fine, since the history is of actual interest in a failure case

PR #1427, opened 2021-03-18
1 group being OK, does not mean that another is the same, so we need …

baldersheim:
…to wait until the cluster has converged.

There will also be lidspace compaction inflight that causes som fluctuations.

@geirst PR


baldersheim:
@geirst ping

geirst (approved):
👍

PR #1434, opened 2021-03-20
Wait for file to exist on all servers

hmusum:
There is a maintainer that downloads missing file references, it
runs every 30 seconds and bails out if it does not get a lock, so
loop and retry for some time before giving up.


aressem (approved):

baldersheim (commented):
👍

PR #1447, opened 2021-03-25
Balder/common proton restart

baldersheim:
@geirst or @toregge PR


toregge (approved):
Consider using weighted set to trigger use of IDocumentWeightAttribute.

baldersheim:
You are right, not intuitive for the commoner that is faster, so I will bring it in as a separate PR

PR #1449, opened 2021-03-26
Use faster wand searcher for testing hash lookup performance.

baldersheim:
iYql is too costly for this use.

@toregge or @geirst PR


geirst (approved):
👍

PR #1453, opened 2021-03-29
try to extract more info about possible hits

arnej27959:
* also, faster feeding of 5k documents

@baldersheim please review
NOTE: this should (in theory) not avoid the issue we've seen, just confirm if my understanding is correct. Will change to starting with 2 searchable copies later, that should avoid the issue.


baldersheim (commented):
👍  Sorry, I missed this one.

PR #1505, opened 2021-04-20
Remove 'none' from expected message

hmusum:
There was no 'none' in expected message before
https://github.com/vespa-engine/system-test/pull/1502 and code seems
to have a newline just before nodes or 'none' is added anyway.

Actual log message found in test run:

Cluster dispatcher.search: group 0 has reduced coverage: Active documents: 0/0, working nodes: 0/1 required 1, unresponsive nodes: 
search node key = 0 hostname = eea08fbd.systemtest.vespa-vespa-factory.gq2.ows.oath.cloud path = 0 in group 0 statusIsKnown = true working = false activeDocs = 0


bratseth (approved):
Thanks

PR #1517, opened 2021-04-27
add screwdriver.yaml

kkraune:


kkraune:
it looks like it is running tests of the test framework itself ;-)

aressem (approved):

PR #1527, opened 2021-04-30
Simplify container http performance test

bjorncs:
Only benchmark async write (which is the whole use case that is actually relevant).
Rework all graphs.
Deduplicate code.


jonmv (approved):
Great!

PR #1568, opened 2021-05-11
Make sure feed files are unique per test

hmusum:
Tests in a class run in parallel, make sure to use a unique generated
file per test (tmpdir includes test method in name, so should be unique)

### DIFF
--- a/tests/search/stableactivecount/stableactivecount.rb
+++ b/tests/search/stableactivecount/stableactivecount.rb
@@ -150,23 +150,25 @@ class StableActiveCount < SearchTest
     puts "Generating feed 1"
     first_pass_docs = 100_000
     content = { :field1 => 'hello', :field2 => '2000' }
+    feed_file_1 = "#{dirs.tmpdir}1stfeed.xml"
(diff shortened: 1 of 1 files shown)


baldersheim (approved):

tests/search/stableactivecount/stableactivecount.rb
```diff
@@ -150,23 +150,25 @@ def run_stable_active_count_test
...
     puts "Feeding first feed"
-    feed(:file => 'generated/1stfeed.xml')
+    feed(:file => feed_file_1)
 
     search_threads = 6
     threads, runners = launch_search_threads(search_threads, first_pass_docs)
 
     puts "Feeding second feed"
-    feed(:file => 'generated/2ndfeed.xml')
+    feed(:file => feed_file_1)
```
geirst, line 171: @hmusum This is the wrong file, should be _feed_file_2_.

hmusum: crap, cut-and-paste error.  will fix

PR #1572, opened 2021-05-12
Dump status page to help understand why resume completes immediately

hakonhall:
Also, test_upgrade_of_downed_node waits until the state is 'down', instead of
one of "sdm" and waiting 30s.

Also, when waiting for the cluster controller to ack a storage node being
stopped, we used to wait until the the node were in one of the states
'stopping', 'maintenance', or 'down'.  However a storage node in 'stopping' is
a transitional state that starts with the node behaving as-if 'up', which the
client probably does not want.  Therefore 'stopping' is removed from this set.

### DIFF
--- a/lib/vespa_model.rb
+++ b/lib/vespa_model.rb
@@ -62,12 +62,12 @@ class VespaModel
     return @storage[cluster].storage[index.to_s]
   end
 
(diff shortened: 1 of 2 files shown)


hakonhall:
FYI @havardpe

vekterli (approved):
:+1:

PR #1578, opened 2021-05-13
Test the same functionality in a simpler way

bratseth:


bratseth:
Ping @arnej27959

arnej27959 (approved):
we should check the "ü" / "ue" case also since that expands from one to two letters?

bratseth:
ugh, yes good point. I'll try to find an example that works in both stemmers.

bratseth:
I'm unable to find any.

PR #1580, opened 2021-05-13
Reduce sensitivity to linguistic differences

bratseth:
There were two problems here:
- Dynamic summaries are sensitive to the length of the text including stems, and therefore to the exact stemming of some piece of text.
- The test relied on wrong/questionable stemming of "blues", and "number"/"numb".


arnej27959 (approved):
LGTM

PR #1583, opened 2021-05-14
Suspend until success

hakonhall:
When actually asking the CC for suspension (https://github.com/vespa-engine/vespa/pull/17854), the suspend of the node with the downed storage node failed due to 409
```
{
  "hostname" : "aa720a43.systemtest.vespa-vespa-factory.gq2.ows.oath.cloud",
  "reason" : {
    "constraint" : "controller-set-node-state",
    "message" : "suspend failed: Changing the state of aa720a43.systemtest.vespa-vespa-factory.gq2.ows.oath.cloud would violate controller-set-node-state: Failure from cluster controllers [c02476e5.systemtest.vespa-vespa-factory.gq2.ows.oath.cloud] when setting node 0 in cluster music to state MAINTENANCE: Task exceeded its version wait deadline: the following nodes have not converged to at least version 6: distributor.0, distributor.1, distributor.2, storage.0, storage.1, storage.2"
  }
}
```

The typical deadline is just short of 10s I believe, and I guess this is too short on the system test nodes!?

This may also happen in real production environment, and in case the host admin will retry.  So the test is changed to retry until success or 10m timeout.

### DIFF
--- a/tests/cloudconfig/orchestrator/orch_cluster.rb
+++ b/tests/cloudconfig/orchestrator/orch_cluster.rb
@@ -243,13 +243,22 @@ class OrchestratorContainerClusterTest < CloudConfigTest
     stop_wait_timeout = 600
     @vespa.stop_content_node("music", 0, stop_wait_timeout, 'd')
 
-    assert_response_code(orch_suspend(@contentD), 409)
-    assert_response_code(orch_suspend(@contentE), 409)
+    # Wait until stopped node is allowed suspension.  The first suspend may not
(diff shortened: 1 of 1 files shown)


hakonhall:
In case there is another reason for this, another mystery opens.  Let me dump the status page just before to get more info.

havardpe (approved):
maybe we want to do all 3 in a loop, just to verify that even when we are eventually allowed to suspend our node, the others are still off limits...

hakonhall:
Done

havardpe (approved):

PR #1594, opened 2021-05-18
Revert "Reduce sensitivity to linguistic differences"

bratseth:
Reverts vespa-engine/system-test#1580

Still differences between stemmers.


arnej27959 (commented):
👍

PR #1608, opened 2021-05-20
Comment out queries that are never verified

hmusum:
Seems these queries are never asserted, not sure why this is run. Logs a lot of data and takes 300 seconds, could it be removed?

@aressem FYI


arnej27959:
it looks like it tries to stress test a plugin and provoke some sort of log message (something that is logged
every 5 minutes), but I don't remember what that was supposed to be.  We should probably remove the entire 
test_friends_bm_plugin method.

hmusum:
OK, test method removed

arnej27959 (approved):
👍

PR #1609, opened 2021-05-20
Arnej/test hex tensor feed

arnej27959:

@geirst please review
@jobergum @lesters FYI

### DIFF
--- a/tests/search/tensor_feed/tensor_feed_hex.rb
+++ b/tests/search/tensor_feed/tensor_feed_hex.rb
@@ -1,0 +1,179 @@
+
+require 'indexed_search_test'
+require 'search/tensor_feed/tensor_feed_base.rb'
+
+class TensorFeedHexTest < IndexedSearchTest
+
+  include TensorFeedTestBase
+
+  def setup
+    set_owner("arnej")
+    @base_dir = selfdir + "tensor_hex/"
+  end
+
+  def test_tensor_hex_feed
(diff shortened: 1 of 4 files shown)


jobergum:
Does it make sense for float as well?

tests/search/tensor_feed/tensor_feed_hex.rb
```diff
@@ -0,0 +1,173 @@
+
```
geirst, line 1: Please add copyright.

tests/search/tensor_feed/tensor_feed_hex.rb
```diff
@@ -0,0 +1,173 @@
...
+  include TensorFeedTestBase
+
+  def setup
+    set_owner("arnej")
+    @base_dir = selfdir + "tensor_hex/"
+  end
+
+  def test_tensor_hex_feed
+    set_description("Test feeding of tensor field and retrieval via search and visit")
```
geirst, line 15:
```suggestion
    set_description("Test hex format feeding of tensor field and retrieval via search and visit")
```

geirst (approved):
👍

arnej27959:
> Does it make sense for float as well?

maybe, maybe not.  Each float would need 8 hex digits, which may be more or less text than normal json floats, depending on how much precision etc they have.  I'm not sure what the difference in cost of parsing would be.

PR #1621, opened 2021-05-28
Bjorncs/vespa feed client

bjorncs:

### DIFF
--- a/lib/document_set.rb
+++ b/lib/document_set.rb
@@ -45,12 +45,11 @@ class DocumentSet
       if first
         first = false
       else
-        f.write(",")
+        f.write(",\n")
       end
       f.write(document.to_put_json(false))
-      f.write("\n")
     end
-    f.write("]\n")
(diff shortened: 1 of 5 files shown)


lib/document_set.rb
```diff
@@ -37,6 +37,22 @@ def write_vespafeed_xml(name)
...
+    @documents.each do |document|
+      if first
+        first = false
+      else
+        f.write(",")
+      end
+      f.write(document.to_put_json(false))
+      f.write("\n")
+    end
+    f.write("]\n")
```
jonmv, line 53:
```suggestion
    f.write("[\n")
    first = true
    @documents.each do |document|
      if first
        first = false
      else
        f.write(",\n")
      end
      f.write(document.to_put_json(false))
    end
    f.write("\n]\n")
```
Put commas at end of line :>

jonmv (commented):
Nice!

PR #1643, opened 2021-06-02
Add initial benchmark for vespa-feed-client programmatic

bjorncs:


tests/performance/programmatic_feed_client/java-feed-client/src/main/java/com/yahoo/vespa/systemtest/javafeedclient/VespaFeedClient.java
```diff
@@ -0,0 +1,65 @@
...
+        String route = System.getProperty("vespa.test.feed.route");
+        int documents = Integer.parseInt(System.getProperty("vespa.test.feed.documents"));
+        AtomicInteger successfulRequests = new AtomicInteger(0);
+        AtomicInteger failedRequests = new AtomicInteger(0);
+        CountDownLatch doneSignal = new CountDownLatch(documents);
+        long start = System.nanoTime();
+        try (FeedClient client = createFeedClient()) {
+            for (int i = 0; i < documents; i++) {
+                DocumentId id = DocumentId.of("music", "music", Integer.toString(i));
+                client.put(id, "{\"fields\": {\"title\": \"Ronny och Ragge\"}}", OperationParameters.empty().route(route))
```
jonmv, line 33: 😂

tests/performance/programmatic_feed_client/java-feed-client/src/main/java/com/yahoo/vespa/systemtest/javafeedclient/VespaFeedClient.java
```diff
@@ -0,0 +1,65 @@
...
+                        .whenComplete((result, error) -> {
+                            if (result != null) {
```
jonmv, line 35: The idiom is to check whether `error != null`, as `result` in general may be `null`. In this case it will never be, though :)

jonmv (approved):
Unicorn

PR #1674, opened 2021-06-08
Real process is +1 of shell pid

jonmv:
@bjorncs


bjorncs (approved):

arnej27959:
it is simpler and more robust to change "java ..." to "exec java ..." and the pid will remain the same.

jonmv:
Thanks @arnej27959 !

PR #1685, opened 2021-06-11
Minor cleanup of bundle stuff and add some debug logging

hmusum:
Trying to find a bug, but unable to reproduce, so add some logging


aressem:
👍  Did not hit the approve button.

PR #1689, opened 2021-06-14
Balder/restart in order to get stable performance

baldersheim:
@havardpe or @geirst PR

### DIFF
--- a/lib/nodetypes/vespa_node.rb
+++ b/lib/nodetypes/vespa_node.rb
@@ -217,6 +217,11 @@ class VespaNode
     logcontent.scan(regexp).size
   end
 
+  def trigger_flush
+    output = `vespa-proton-cmd --local triggerFlush 2>&1`.chomp
+    testcase.output(output)
+  end
+
   # The time on this Vespa node.
   def time
     Time.new
--- a/tests/performance/nearest_neighbor/nearest_neighbor.rb
+++ b/tests/performance/nearest_neighbor/nearest_neighbor.rb
@@ -100,8 +100,8 @@ class NearestNeighborPerformanceTest < PerformanceTest
     wait_for_hitcount("sddocname:foobar", 100000, 30)
     puts "DONE FEEDING"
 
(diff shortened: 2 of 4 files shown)


havardpe (approved):

tests/performance/nearest_neighbor/nearest_neighbor.rb
```diff
@@ -100,8 +100,8 @@ def test_nearest_neighbor
     wait_for_hitcount("sddocname:foobar", 100000, 30)
     puts "DONE FEEDING"
 
-    node.execute('vespa-proton-cmd --local triggerFlush')
-    node.execute('vespa-proton-cmd --local triggerFlush')
+    node.trigger_flush
```
geirst, line 103: Consider using _vespa.search["search"].first.trigger_flush_ instead.

lib/nodetypes/vespa_node.rb
```diff
@@ -217,6 +217,11 @@ def log_matches(regexp)
     logcontent.scan(regexp).size
   end
 
+  def trigger_flush
```
geirst, line 220: I am not sure we want a _trigger_flush_ in this class, as it only works when its actually a SearchNode on the same host. nodetypes/searchnode.rb already has this function defined.

tests/performance/phrase/phrase_cases.rb
```diff
@@ -64,9 +64,8 @@ def test_phrase_cases
     (exitcode, output) = execute(node, "vespa-feed-perf < #{dirs.tmpdir}/feed-phrases.xml")
     assert_equal(0, exitcode)
     wait_for_hitcount("sddocname:foobar", 123456, 30)
-
-    node.execute('vespa-proton-cmd --local triggerFlush')
-    node.execute('vespa-proton-cmd --local triggerFlush')
+    node.trigger_flush
```
geirst, line 67: Consider using _vespa.search["search"].first.trigger_flush_ instead. Also consider adding the die function below to nodetypes/searchnode.rb.

PR #1690, opened 2021-06-14
Rewrite test to deploy and check file distribution in separate threads

hmusum:
Need to do this in 2 threads to measure file distribution time correctly

Test needs 2 nodes and haven't been able to run this successfully in multi-node setup, so this change might need some iterations.

@geirst please review


aressem (approved):

aressem:
Remember that Ruby has the global instruction lock so they are not executed simultaneously. It depends on what he functions called in the threads do.

hmusum:
OK, let's see what happens

PR #1715, opened 2021-06-15
Generate feed with c++ code instead of slow ruby.

baldersheim:
@hmusum or @vekterli PR

### DIFF
--- a/tests/performance/collection_feed/weighted_set_feed.rb
+++ b/tests/performance/collection_feed/weighted_set_feed.rb
@@ -23,20 +23,6 @@ class WeightedSetFeedTest < PerformanceTest
     super
   end
 
-  def with_json_feed(file_name)
-    raise 'No block given' if not block_given?
-    json_writer = JsonDocumentWriter.new(File.open(file_name, 'w'))
-    begin
-      yield json_writer
-    ensure
-      json_writer.close()
-    end
-  end
-
-  def doc_id(n)
-    "id:user:footype::#{n}"
-  end
-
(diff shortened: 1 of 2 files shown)


hmusum (approved):

tests/performance/collection_feed/weighted_set_feed.rb
```diff
@@ -139,11 +118,15 @@ def test_wset_attribute_feed_performance
...
+    @data_generator = "#{dirs.tmpdir}/docs"
+    container = (vespa.qrserver["0"] or vespa.container.values.first)
+    container.execute("g++ -Wl,-rpath,#{Environment.instance.vespa_home}/lib64/ -g -O3 -o #{@data_generator} #{selfdir}/docs.cpp")
+
     doc_count = 10_000_000
     parameter_combinations.each do |p|
       # Reduce document count for large cardinalities to keep test time reasonable.
       test_doc_count = doc_count / p.wset_size
-      test_doc_count *= 5 if !p.fast_search
+      test_doc_count *= 2 if !p.fast_search
```
vekterli, line 129: Is it necessary to reduce the doc count now that we have a fast doc generator?

vekterli (commented):
:+1: should we have some test framework functionality to compile native stuff as part of the test? Or do we already have this squirreled away in some dark corner of the code base already?

tests/performance/collection_feed/weighted_set_feed.rb
vekterli: Disregard, I just noticed the other PR that added the code in the first place... 🙂

PR #1721, opened 2021-06-16
It is not possible to make this assert deterministic, so just remove …

baldersheim:
…the conditional assert alltogether.

@geirst PR


geirst (approved):
👍

PR #1759, opened 2021-06-21
Stop using unsupported tuning option

hmusum:
@geirst or @baldersheim please review and merge, but only if we are sure that nobody is using this option. If someone are using it, next release will be broken.


hmusum:
@aressem FYI

baldersheim (approved):

baldersheim:
It has never been used, it was removed from doc 3 years ago.

PR #1816, opened 2021-07-06
Reduce slack for max hits. Worst outliers are now ignored (due to them

toregge:
being caused by slow poll results).

Reduce slack for max docs for same reason.

@geirst : please review


geirst (approved):
👍

PR #1818, opened 2021-07-07
Reduce from 46Gb of feed data written to disk

jonmv:
@bjorncs please mereg.


jonmv:
We're killing the opensource test, as the host disk goes full >_<

bjorncs (approved):
😄

PR #1821, opened 2021-07-08
Debug result parsing

hmusum:
Fails occasionally on opensource run


jonmv (approved):
Used to fail because disk got full. Yes, we need to see this ...

PR #1845, opened 2021-08-20
Change ownership of generated config file to original user

bjorncs:

### DIFF
--- a/lib/tls_env.rb
+++ b/lib/tls_env.rb
@@ -61,6 +61,7 @@ class TlsEnv
       File.open(tls_config_file, 'w') do |f|
         f.syswrite(json)
       end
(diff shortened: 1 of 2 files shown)


lib/tls_env.rb
```diff
@@ -61,6 +61,7 @@ def generate_tls_config_if_missing
       File.open(tls_config_file, 'w') do |f|
         f.syswrite(json)
       end
+      FileUtils.chown(ssl_config.user, nil, tls_config_file)
```
aressem, line 64: If this file is written into /root/.vespa/.../tls_config.json, this will not allow the config server running as vespa to read this file? I this tested now to work in the PR build?

bjorncs: You are right - that needs to be resolved as well.

aressem (approved):

PR #1863, opened 2021-09-15
Wait until content nodes have new config before calling wait_until_ready

hmusum:
These tests became unstable when doing some config protocol changes (that probably changed how fast config was updated for nodes). 

Taken from tests/vds/join_sparse_buckets/join_sparse_buckets.rb, which probably had the same issue at some time


baldersheim (approved):

vekterli (commented):
:+1:

PR #1864, opened 2021-09-15
Wait until content nodes have new config before calling wait_until_ready

hmusum:
One more test that need to wait, see https://github.com/vespa-engine/system-test/pull/1863


vekterli (approved):
:+1:

PR #1870, opened 2021-09-17
Always validate cluster bucket state

hmusum:
Always validate, remove third argument to method, as it is never used.


vekterli (approved):
:+1:

PR #1873, opened 2021-09-20
Only break out of loop if response code is 200

hmusum:
Simplify some timing code and remove some logging

Seeing some tests that seem to fail with no data in response body, maybe this will help...


vekterli (approved):
:+1:

PR #1888, opened 2021-09-27
Update expected error message

bratseth:
Isn't checking specific error messages better suited for unit testing?


hmusum (approved):
Yes, we can probably just check that it fails and add to some unit tests if not already covered by existing tests. I can have a look.

PR #1912, opened 2021-10-07
Update expected md5sums

hmusum:
Seems like changing copyright altered the checksum (I thought comments
should be ignored when calculating checksums, though...)


hmusum:
The checksums are generated by the test by doing `md5sum foo.def` and does not remove any comment lines, so this is an "issue" with the test. I think we can live with this.

PR #1916, opened 2021-10-08
Generate feed up front, then cat

jonmv:
@geirst please review. Let's consider this if it makes tests more stable, and doesn't cost too much time/space.


jonmv:
Time saving are massive over downloading the data, anyway :)

jonmv:
This had no stabilising effect. It's rather the feed format that is to blame. This did increase the throughput of some of the feeding tests, though.

I any case, it's simpler to pipe straight to the feeder process. It's either that, or pipe to the feeder to store in .vespa format temporarily. Since the usual setup is client on a different node than server, it makes sense to reduce the client impact as much as possible. On the other hand, feeding over HTTP uses .json which is parsed inside the container, anyway. I suggest we revert this for now.

PR #1942, opened 2021-10-21
Split docker file in base and systemtest because the docker versions …

aressem:
…we use can't skip build steps based on build arguments.


aressem:
Review only. Must be coordinated with https://github.com/vespa-engine/vespa/pull/19624

hmusum (approved):

PR #1975, opened 2021-10-30
Reduce max concurrency

jonmv:
@bjorncs please review. I wonder if there is simply a problem with 64 concurrency ruby executor tasks on centos?


jonmv:
Or @freva :)

freva (approved):

PR #2046, opened 2021-12-01
Make sure to return deploy output from deploy_app_and_assert

hmusum:
Also no need to use default value as seconds argument to wait_for_reconfig

Fix after changes in https://github.com/vespa-engine/system-test/pull/2045


hmusum:
ping @baldersheim

baldersheim (approved):

PR #2056, opened 2021-12-09
Wait for all services to be reconfigured

hmusum:
wait_for_application waits for app checksum, which is unchanged,
so will have no effect in this case.


bjorncs (approved):
👍

PR #2078, opened 2021-12-16
Actually compare fields.

toregge:
@baldersheim : please review


baldersheim (approved):
Will there be havoc ....

PR #2101, opened 2022-01-05
Use 250 instead of 200 as slack_minhits

hmusum:
Tyring to stabilize test ...


baldersheim (dismissed):
This parameter is not set in the GrowApp object already, and should have been set there instead ?

hmusum:
Maybe that's better, I'll have a look

hmusum:
@baldersheim please take a look again

baldersheim (approved):

PR #2103, opened 2022-01-05
Revert "Remove 'operation="append"' for config overrides."

hmusum:
Reverts vespa-engine/system-test#2102

Looks like this broke some system tests, for example GenericConfig::test_generic_config (tests/search/generic_config/generic_config.rb)


aressem (approved):

gjoranv:
Thanks, my usage search was faulty.

PR #2141, opened 2022-01-26
Try to fix copying of files

hmusum:


tests/performance/onnxmodel/onnxmodel.rb
```diff
@@ -40,7 +40,7 @@ def test_onnx_deploy
   def fetch_model
     @node = vespa.nodeproxies.first[1]
     file = @node.fetchfiles(:webhost => WEBHOST, :file => "pub/systemtests/performance/onnxmodel/ranking_model.onnx", :nocache => true, :nochecksum => true).first
```
baldersheim, line 42: But can this test be here ? or is WEBHOST public ?

hmusum: I think you are right, this needs to be an internal test

hmusum: Or we could generate onnx file, but that would require some dependencies to be installed, need to look into that

jobergum: It would be great to have this in the public - we can pull the model from our public S3 bucket?

hmusum: Done

hmusum:
@jobergum or @baldersheim please take a look

jobergum (approved):

PR #2146, opened 2022-01-27
Remove app used with yourkit that does not work anymore

hmusum:
`<docproc>` is not supported any longer, will remove framework
code for this in another PR, as it does not work and is unused

@geirst please review, looks like this has not been run with YourKit in a long time, as app does not work on Vespa 7. Could be rewritten, but would rather do it only if needed


geirst (approved):
👍

PR #2150, opened 2022-01-27
Wait for bucket replica readiness state propagation

vekterli:
@geirst please review. This is expected to resolve the rare instabilities of this test.

Replica readiness updates are communicated and propagated asynchronously
from Proton via the persistence layer and finally to the distributor.
To avoid spurious failures where we consistency-check the bucket DBs
prior to readiness changes being propagated, ensure we retry in the face
of transient (but expected) inconsistencies.

Timeout is long enough that even Valgrind runs should be able to keep up.


geirst (approved):
👍

PR #2154, opened 2022-01-28
GC dead code

hmusum:
'docproc' and 'search' has not been used as top-level elements
in services.xml for many years.


baldersheim (approved):
👍  Seems like the jvm_args stuff has been a bit of a mess :)

PR #2162, opened 2022-01-31
Bratseth/rankprofile modularity

bratseth:
I'm hijacking this test to test rankprofile modularity, it has some small value as it is so that seems better than removing it.

This PR should have it working just as before, I'll follow up with one adding test of the new modularity + multiple inheritance.


geirst (dismissed):
👍  Should probably rename the test class (RankExpressionLogical) as well.

bratseth:
Missed that, thanks

PR #2166, opened 2022-01-31
Revert "Modularize rank-profiles"

hmusum:
Reverts vespa-engine/system-test#2164

System test started failing with this change, not obvious to me what's wrong here (test changes or new code for using modular rank profiles).


freva (approved):

bjormel (approved):

bratseth:
I annotated the test and will follow up.

PR #2173, opened 2022-02-02
Add container with document api when 'enable_document_api' is used

hmusum:
Cleanup in next PR (remove code in `Clients`)

PR #2174, opened 2022-02-03
Revert "Add container with document api when 'enable_document_api' is used"

hmusum:
Reverts vespa-engine/system-test#2173

Looks like this broke some system tests, need to investigate

### DIFF
--- a/lib/app_generator/app.rb
+++ b/lib/app_generator/app.rb
@@ -124,11 +124,8 @@ class App
     return self
   end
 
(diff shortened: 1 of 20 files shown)


hmusum:
Looks like `wait_for_hitcount` sends query to the first container and `doc-api` container now comes before default (search) container if both are defined.

gjoranv (approved):

bratseth (approved):

PR #2177, opened 2022-02-04
decode summary features also when parsing XML

arnej27959:

@baldersheim 
note: this is one small step towards flipping the default in system tests from format=xml to json


baldersheim (approved):
A giant step :)

PR #2187, opened 2022-02-07
GC dead code

hmusum:
No document api container, so feeder option is not used.

### DIFF
--- a/tests/search/grouping/grouping.rb
+++ b/tests/search/grouping/grouping.rb
@@ -1,4 +1,4 @@
-# Copyright 2019 Oath Inc. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+# Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 require 'indexed_search_test'
 
 class Grouping < IndexedSearchTest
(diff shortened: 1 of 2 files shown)


tests/search/grouping/grouping.rb
```diff
@@ -7,8 +7,7 @@ class Grouping < IndexedSearchTest
 
   def setup
     set_owner("bjorncs")
-    deploy_app(SearchApp.new.sd("#{selfdir}/purchase.sd").
-               feeder_options(FeederOptions.new.max_pending_docs(1)))
+    deploy_app(SearchApp.new.sd("#{selfdir}/purchase.sd"))
```
baldersheim, line 10: I wonder if max_pending=1 is required to get a deterministic test.
Is there another way to achieve that, or does this mean that it has not been working ?

baldersheim (approved):

tests/search/grouping/grouping.rb
hmusum: This config has not been working but there is config for the feeder in the test that will do what is wanted I think:

`    feed_and_wait_for_docs("purchase", 20, :file => "#{selfdir}/docs.xml", :maxpending => 1);`

baldersheim: Perfect, just leftovers then.

PR #2221, opened 2022-02-17
Change the way copying of downloaded file is done

hmusum:
I have never gotten downloading a file and copying to test dir work at first attempts. It might be that it does not work either, can't find any examples on how this should be done.


hmusum:
@aressem please take a look

aressem (approved):

PR #2247, opened 2022-03-25
handle embedded nil/false in structures

arnej27959:
* do quick compare with .equal? first
* check for existence with .has_key?

@geirst or @jonmv please review


geirst (approved):
👍 Consider adding new test case(s) as well.

PR #2252, opened 2022-03-28
use "mysearch" as source/cluster name

arnej27959:

@bratseth or @aressem please review and merge


bratseth (approved):
👍   "my*" is the only name scheme I've found that doesn't end up in customer applications in production.

PR #2295, opened 2022-04-27
Add inputs

bratseth:


arnej27959 (approved):
for some reason i expected
  `query(qi) : tensor(x[2])`
but i guess you decided it was simpler without any colon.
👍

bratseth:
Somebody pointed out we might want to support setting default values, which would make this
query(qi)  tensor(x[2]):[1.2]

PR #2307, opened 2022-05-02
Add system test for using boolean fields in visitor selections

vekterli:
@geirst please review

Tests attribute fields, non-attribute fields and combinations thereof.


geirst (approved):
👍

PR #2309, opened 2022-05-02
Reduce spam-level logging in unstable test

vekterli:
@geirst please review

Too much logging makes test stable, which paradoxically is the exact
opposite of what we want to achieve...!


geirst (approved):
👍

PR #2323, opened 2022-05-12
Run visiting for twice as long, see if it stabilizes performance

hmusum:


hmusum:
test runs on rhel8 seems to have gotten more unstable, centos7 unchanged 🤷

PR #2329, opened 2022-05-16
test default query feature tensor value

havardpe:
@geirst please review
@bratseth FYI

depends on https://github.com/vespa-engine/vespa/pull/22618 to unbreak tensor query features for streaming search


geirst (approved):
👍

PR #2350, opened 2022-05-30
Use modern syntax

bratseth:


baldersheim (approved):
622..., took some serious scrolling time

PR #2372, opened 2022-06-03
Use the default summary class as the minimal class doesn't exist.

geirst:
On Vespa 8 it is not allowed to specify a non-existing summary class.
We also ended up using the default summary class before, as a fallback.

@arnej27959 please review

### DIFF
--- a/tests/performance/wand_performance/elasticspapp/schemas/test.sd
+++ b/tests/performance/wand_performance/elasticspapp/schemas/test.sd
@@ -13,6 +13,9 @@ schema test {
         }
     
     }
+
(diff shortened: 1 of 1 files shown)


arnej27959:
we can try this, but old behavior was no summary data from backend AFAIK; equivalent would be adding
`document-summary minimal {}`
in .sd file

geirst:
Ah, I see @arnej27959. Then I will add that instead.

geirst:
@arnej27959 PTAL

arnej27959 (approved):
👍

PR #2400, opened 2022-06-07
Update to 8-SNAPSHOT and 8.*.0

gjoranv:
Please verify the change in `bin/run-tests-on-swarm.sh`

### DIFF
--- a/bin/run-tests-on-swarm.sh
+++ b/bin/run-tests-on-swarm.sh
@@ -228,6 +228,8 @@ VESPAVERSION=$(docker run --rm $BINDMOUNT_OPTS --entrypoint bash $DOCKERIMAGE -l
 case "$VESPAVERSION" in
     7.*.0) VESPAVERSION=7-SNAPSHOT
 	   ;;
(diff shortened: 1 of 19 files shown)


bin/run-tests-on-swarm.sh
```diff
@@ -226,7 +226,7 @@ fi
 
 VESPAVERSION=$(docker run --rm $BINDMOUNT_OPTS --entrypoint bash $DOCKERIMAGE -lc '${VESPA_HOME-/opt/vespa}/bin/vespa-print-default version')
 case "$VESPAVERSION" in
-    7.*.0) VESPAVERSION=7-SNAPSHOT
+    8.*.0) VESPAVERSION=8-SNAPSHOT
```
aressem, line 229: Just leave both cases there so it works for 7 and 8 in a limited time period.

aressem (changes requested):

bin/run-tests-on-swarm.sh
gjoranv: Added 7.*.0 back

PR #2417, opened 2022-06-10
Update test after weakAnd became default query type

hmusum:
Add weakand result file for default query
Add &type=all for some queries, remove where other type was already specified


bratseth (approved):
👍

PR #2437, opened 2022-06-27
Add timeout to visit tests

jonmv:
@freva


jonmv:
Without this, a visit would run for up to 60s ...

freva (approved):

PR #2444, opened 2022-07-01
Check empty fields in response also for streaming search.

toregge:
@geirst : please review
@baldersheim : FYI

This PR depends on vespa-engine/vespa#23315


geirst (approved):
👍

PR #2452, opened 2022-07-04
Fix syntax error in tests.

aressem:


baldersheim (approved):
Uh, tests have failed without notice ?

aressem:
Test files that that are uninterpretable or does not contain any test methods are discarded like they have always been.

PR #2483, opened 2022-07-11
Always feed "boool" field (which is of type bool) to allow expected

toregge:
answers to be shared between tests for indexed and streaming mode
when bool fields are not stored in docsum blob for streaming search.

@geirst : please review


geirst (approved):
👍

PR #2484, opened 2022-07-11
Unset bool field is not printed for streaming search when

toregge:
bool values are not stored in docsum blob.

This should be merged at same time as vespa-engine/vespa#23447

@geirst : please review


geirst (approved):
👍

PR #2490, opened 2022-07-19
Add support for fetching testdata from S3.

aressem:

### DIFF
--- a/lib/remote_file_utils.rb
+++ b/lib/remote_file_utils.rb
@@ -1,0 +1,27 @@
+
+require 'uri'
+require 'fileutils'
+
+module RemoteFileUtils
+
+  # Fetch a single URL to a local file destination
+  # Supports http(s) and s3 using wget and aws cli
+  def download(source_url, localfile)
+    FileUtils.mkdir_p(File.dirname(localfile))
+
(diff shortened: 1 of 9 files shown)


lib/remote_file_utils.rb
```diff
@@ -0,0 +1,27 @@
...
+module RemoteFileUtils
+
+  # Fetch a single URL to a local file destination
+  # Supports http(s) and s3 using wget and aws cli
+  def download(source_url, localfile)
+    FileUtils.mkdir_p(File.dirname(localfile))
+
+    case source_url.scheme
+    when "s3"
+      cmd = "aws s3 cp #{source_url} #{localfile}"
```
bjorncs, line 14: Consider using `aws s3 sync` which will only re-download file if changed

bjorncs (approved):

lib/remote_file_utils.rb
aressem: Will that work if the local file path is different than the S3 object name ?

aressem: I will look into this when we start to use this functionality on AWS, but `sync` seems to work on directories and not single files.

PR #2514, opened 2022-09-05
Test matched elements filter and attribute combiner with no matched e…

toregge:
…lements

(array and weighted set of primitive types).

@geirst : please review


geirst (approved):
👍

PR #2519, opened 2022-09-08
Don't render field in search result if it is empty after matched elem…

toregge:
…ents filtering.

This PR adjusts expected results to match changes in vespa-engine/vespa#23989

@geirst : please review


baldersheim (approved):

geirst (approved):
👍

PR #2522, opened 2022-09-09
Reduce number of variants in test

hmusum:
I think these can be removed without losing much confidence in performance for this feature.


baldersheim (approved):

vekterli (commented):
:+1:

PR #2537, opened 2022-09-16
Removing the perl stuff also removed the openssl dependencies. Use ve…

aressem:
…spa-openssl to generate the certs.

This should fix the build failures for open source on factory.


arnej27959 (approved):
👍

PR #2553, opened 2022-09-23
Reapply "Aressem/retry test if node server dies""

aressem:
Reverts vespa-engine/system-test#2552


aressem:
The observed failures due to this has other causes.

hmusum (approved):

PR #2555, opened 2022-09-26
stop generating RSA keys

arnej27959:
* generate Elliptic Curve keys with P-256 parameter set, like the vespa CLI tool does.

note: this assumes "curl compiled with NSS" is no longer a concern. 

@bjorncs please review
@aressem FYI


bjorncs (approved):

bjorncs:
FYI @vekterli

PR #2581, opened 2022-10-25
Test regexp with prefix matching a single dictionary entry.

toregge:
@geirst : please review

This is related to vespa-engine/vespa#24476


geirst (approved):
👍

PR #2627, opened 2022-11-21
Update expected feature value MERGEOK

baldersheim:
@bratseth or @ean or @frodelu PR


ean (approved):

bratseth:
I think we should test equality instead.

baldersheim:
Not sure I understand...

bratseth:
The purpose of this is to enable comparing strings in ranking expressions, not to output certain numbers, so it it would be better to write e.g a function that compared the attribute with a hardcoded string and check for the value of that function.

PR #2636, opened 2022-11-28
Create docs and feed instead of using xml files

hmusum:


vekterli (approved):
:+1: thus ends the long and rich era of `banana.xml` 🫡

PR #2662, opened 2023-01-02
Use search with search path to poll for document redistribution.

toregge:
@geirst : please review


baldersheim (approved):

toregge:
Note: we still might need a small sleep after polling for document count.

baldersheim:
Time will show.....

PR #2663, opened 2023-01-03
Propagate the perf recording option from testrunner.

aressem:


aressem:
Needed because the testrunner.rb that runs the whole test suite does not use the autorunner.rb that this was wired into.

baldersheim (approved):

PR #2690, opened 2023-01-12
Revert "Make sure that all containers are running with perfmap to get Java sy…"

baldersheim:
Reverts vespa-engine/system-test#2689
@aressem @hmusum or @arnej27959 
Revert this one to see if that is reason for reduced/unstable perfomance and some core dumps in perf tests

### DIFF
--- a/lib/performance_test.rb
+++ b/lib/performance_test.rb
@@ -273,7 +273,6 @@ class PerformanceTest < TestCase
   end
 
   def setup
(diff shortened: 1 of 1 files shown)


hmusum (approved):

aressem:
If this impacts performance we should probably only add the `libperfmap` agent when perf profiling is turned on.

PR #2709, opened 2023-01-18
prepare for match profiling

havardpe:
use tag to find profiling results rather than array index

@geirst please review


geirst (approved):
👍

PR #2724, opened 2023-01-25
Revert "Yamas cleanup 1"

hmusum:
Reverts vespa-engine/system-test#2719@

Still used, it seems...


arnej27959 (approved):
metrixproxy-client still adds "OK|" in front, it seems.

PR #2751, opened 2023-02-27
Restart when changing indexing mode

hmusum:
Changing indexing mode leads to search nodes getting new config ids. Since they might get new config for old config id before being restarted they might get wrong/empty config, which leads to unstabe test.


jonmv (approved):

jonmv:
Hah, thanks!

PR #2780, opened 2023-03-22
Verify that raw values are used for sorting instead of the

toregge:
base64 encoded string values derived from the raw values.

@geirst : please review.


geirst (approved):
👍

PR #2782, opened 2023-03-23
Install zip, needed to build vespa-factory-cli

hmusum:


hmusum:
Fixed in https://git.ouryahoo.com/vespa/systemtests/pull/6188

PR #2789, opened 2023-03-29
Track document meta store no longer explorable as attribute vector.

toregge:
@geirst : please review


geirst (approved):
👍  Exploring of documentmetastore is still available on the documentdb level.

PR #2790, opened 2023-03-29
Support dumping distributor bucket DB on consistency failure

vekterli:
@geirst please review

Debugging aid for one test that on rare occasions appears to not sufficiently wait for maintenance ops to be completed, causing spurious consistency check failures due to running prior to the cluster fully settling.

Now lets a test specify that the state of the bucket DB on a distributor should be dumped on failure, allowing for a manual inspection and comparison.


geirst (approved):
👍

PR #2796, opened 2023-04-11
Add an implicit barrier for buckets pending background indexing

vekterli:
@geirst please review

This helps avoid a race condition in tests where the test framework erroneously believes all maintenance ops are done since the distributor does not _directly_ indicate any buckets have pending maintenance ops.

Count the number of bucket replicas that _should_ (eventually) be indexed and made searchable according to the ideal state, but where this currently is not the case. This works as an implicit barrier to wait for search nodes to asynchronously complete background bucket moving and search node to distributor bucket info change notifications.

This only works reliably as a barrier for non-grouped clusters, as this code does not have enough knowledge of the group topology and ideal state to make an informed decision on what _other_ replicas than the first one should be ready.

### DIFF
--- a/lib/nodetypes/distributor.rb
+++ b/lib/nodetypes/distributor.rb
@@ -23,6 +23,9 @@ class Distributor < VDSNode
         count = count + 1
       end
     }
+    if count == 0
+      count = count_ideal_replicas_not_ready()
+    end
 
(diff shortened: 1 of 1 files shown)


geirst (approved):
👍

PR #2802, opened 2023-04-13
Expect one less since we have added the default value as document 0. MERGEOK

baldersheim:
@geirst or @toregge or @bratseth or @bjorncs PR


toregge (commented):
👍

PR #2827, opened 2023-04-24
old security-env is going away, drop useless check

baldersheim:
@arnej27959 PR
I wonder if this one has been forgotten .....


arnej27959 (approved):
👍

PR #2828, opened 2023-04-24
Add some more information about current state if not in sync.

baldersheim:
@vekterli PR

### DIFF
--- a/lib/nodetypes/distributor.rb
+++ b/lib/nodetypes/distributor.rb
@@ -11,10 +11,16 @@ class Distributor < VDSNode
   end
 
   def is_synced?
+    state = get_sync_state
+    return state == nil
+  end
+
+  def get_sync_state
+    result = []
     statuspage = get_status_page("/distributor?page=buckets")
     if !has_expected_bucket_db_prologue(statuspage)
       testcase.output('NOTE: bucket DB status page did not have expected format, retrying...')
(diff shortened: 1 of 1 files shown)


lib/nodetypes/distributor.rb
```diff
@@ -47,13 +58,17 @@ def is_synced?
...
+
+  def get_ideal_replicas_not_ready
+    non_ready = []
     each_database_bucket do |space, bucket_id, state, raw_state|
       # Output is in ideal state order, so expect the 1st one to always be ready.
       # This should generalize to always be true for indexed, store-only, flat and grouped setups.
       # We'd ideally check the active-flag as well, but clusters without indexed docs will
       # not activate any replicas.
-      non_ready += 1 if not state[0].ready
+      non_ready.push(bucket_id, state) if not state[0].ready
```
vekterli, line 71: Consider pushing as an array "tuple" `[bucket_id, state]` instead to have each entry take up one array slot. Avoids having to always remember to divide by the expected number of entries per non-ready bucket, and makes it easier to associate each bucket with its state.

vekterli (dismissed):
:+1:

lib/nodetypes/distributor.rb
baldersheim: Fixed

PR #2831, opened 2023-04-25
Revert "Only add streaming.selection parameter when in streaming mode."

toregge:
Reverts vespa-engine/system-test#2830

Many system tests broke due to is_streaming not being a known member function.


toregge:
@aressem : FYI

aressem (approved):

PR #2835, opened 2023-05-03
Base64 encoding with padding.

baldersheim:
@bjorncs or @bratseth PR


baldersheim:
Required for https://github.com/vespa-engine/vespa/issues/26947

PR #2849, opened 2023-05-11
Add tests for conditional Remove with inconsistent replicas

vekterli:
@havardpe please review

All the new tests will fail at the moment, as write-repair is not yet wired into conditional Remove operations.

Also factor out some common bits and simplify precondition testing with some not-too-subtle inspiration from the Doc V1 API tests.

### DIFF
--- a/tests/search/inconsistent_replicas/inconsistent_conditional_remove.rb
+++ b/tests/search/inconsistent_replicas/inconsistent_conditional_remove.rb
@@ -1,0 +1,33 @@
+# Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+require_relative 'inconsistent_buckets_base'
+
+class InconsistentConditionalRemoveTest < InconsistentBucketsBase
+
+  def setup
+    set_owner('vekterli')
+    super
+  end
+
+  def remove_with_condition(condition)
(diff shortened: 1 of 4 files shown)


tests/search/inconsistent_replicas/inconsistent_conditional_remove.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  def test_conditional_remove_mismatching_existing_doc_is_not_applied
+    set_description('Test that conditional Remove is NOT applied when document versions are inconsistent ' +
+                    'across replicas and the condition _does not match_ the newest document')
+    make_replicas_inconsistent_for_single_document
+    assert_precondition_failure {
+      remove_with_condition('music.title == "first title"')
+    }
+    verify_document_has_expected_contents(title: 'second title') # Just checks newest version
```
havardpe, line 30: should check all versions to make sure we have not removed the document that actually matched the precondition

vekterli: (discussed offline, summarized here): if the conditional remove goes through on any node, there will exist a tombstone with a higher timestamp than the removed document itself. If that tombstone is present on any node, it will be authoritative and no document will be returned. I.e. this property is implicitly tested by just looking at the newest document version.

PR #2850, opened 2023-05-11
Add streaming selection query parameter in search_base member function.

toregge:
@geirst : please review
@baldersheim : FYI


geirst (approved):
👍

PR #2851, opened 2023-05-11
Remove unneeded streaming.selection=true query parameter, it will be …

toregge:
…added in search_base when running tests in streaming search mode.

@geirst : please review


geirst (approved):
👍

PR #2883, opened 2023-06-02
Revert "Handle possible search handler warmup query in backend metric verific…"

hmusum:
Reverts vespa-engine/system-test#2880

Almost all system tests fail


jonmv:
Urgh.

PR #2884, opened 2023-06-02
Jonmv/account for search handler warmup query

jonmv:
@hmusum or @geirst please review.


hmusum (approved):

hmusum:
Seems like tests/search/metrics/metrics.rb has become unstable. If the query "may or may not hit the backend" I guess that is inevitable. When test fails it fails on line 69, checking a backend metric

jonmv:
Yes: https://github.com/vespa-engine/system-test/pull/2885

PR #2917, opened 2023-06-21
Avoid creating the dummy service each time we create a client. This r…

aressem:
…educes the number of ports and drb threads dramatically.


hmusum (approved):
Nice!

PR #2924, opened 2023-06-28
value is now only grouped for the bound key

havardpe:
@baldersheim please review

Note: change depends on separate PR


havardpe:
depends on actual code change:
https://github.com/vespa-engine/vespa/pull/27568

baldersheim (approved):

PR #2944, opened 2023-07-11
Revert "Run each test case in a separate process."

aressem:
Reverts vespa-engine/system-test#2943

Merge in case of failures.


hmusum (approved):

aressem:
Not needed.

PR #2946, opened 2023-07-11
Use proper yql syntax for range search.

toregge:
@vekterli : please review


vekterli (approved):
:+1: Future @geirst can optionally take a look as well

vekterli (approved):

PR #2956, opened 2023-07-14
No need to skip testing of weighted_set<long> field w2

baldersheim:
@toregge PR
Once https://github.com/vespa-engine/vespa/pull/27765 is in this one should go in too.


toregge (approved):
:+1:  @vekterli: FYI

PR #2966, opened 2023-07-21
Use the builtin perf map dump in JDK 17

aressem:


aressem:
This enables us to get rid of the custom perf_map_agent libraries.

hmusum (approved):
🏅

PR #3003, opened 2023-08-30
Use stor-communicationmnager config instead of old config that is lon…

baldersheim:
…g gone.

@vekterli or @hmusum PR


hmusum:
👍

vekterli (commented):
:+1:

PR #3011, opened 2023-08-31
Revert "Turn on reconfigurable dispatcher for factory"

jonmv:
Reverts vespa-engine/system-test#3010

@freva or @baldersheim 

I'm not surprised if something breaks.


jonmv:
At least one test broke, and I need to fix something.

PR #3046, opened 2023-09-27
Revert "Fail when using user config with unknown config definition"

hmusum:
Some failures, need to analyze/fix them before we can use 'fail'


aressem (approved):

hmusum:
Not needed

PR #3050, opened 2023-10-09
Don't use triggerFlush after enabling interleaved features. Instead,

toregge:
poll state/v1 api to check for pending urgent flush.

@geirst : please review


geirst (approved):
👍

PR #3066, opened 2023-10-13
Revert "Enable distributor operation cancellation on factory"

hmusum:
Reverts vespa-engine/system-test#3063

Looks like at least one performance test fails (core dump in distributor)


aressem (approved):

vekterli (commented):
:+1:

PR #3070, opened 2023-10-16
Use FullParams

hmusum:
@vekterli or @aressem Use existing class, see https://github.com/vespa-engine/system-test/pull/3069


vekterli (approved):
:+1: d'oh!

PR #3071, opened 2023-10-16
Explicitly use OpenNlpLinguistics to get the same results

toregge:
between public and internal system test runs.

@geirst : please review


geirst (approved):
👍

PR #3077, opened 2023-10-19
test for bug causing container startup failure

arnej27959:
@havardpe please review
@bjorncs @bjormel @baldersheim FYI


havardpe (approved):
consider adding test actually using it (not just start-up test) as a second step

PR #3083, opened 2023-10-25
Remove old debug logging statements

vekterli:
@geirst please review

I am sure these were useful at some point, but not these days 🧹 🦖


geirst (approved):
👍

PR #3110, opened 2023-11-08
Reapply "Hide stderr by default, use :stderr => true if needed."

hmusum:
Reverts vespa-engine/system-test#3104

FYI: Work in progress @baldersheim


hmusum:
Might break something, so review only

baldersheim (approved):

PR #3118, opened 2023-11-15
Run Vespa processes with `no_new_privs` bit set

vekterli:
@aressem please review (and let me know if this kind of stuff should reside somewhere else in the node server).

It's not expected that this should affect anything, but this is one of the least exciting ways to find out.


aressem:
I think this is ok as long as it should not be possible to override.

aressem (approved):

PR #3130, opened 2023-11-22
Revert "Revert "Get stderr (ZK logging) and try 3.9.1 again""

jonmv:
Reverts vespa-engine/system-test#3128

@bjorncs please review, Will try this with https://github.com/vespa-engine/system-test/pull/3129.


jonmv:
I noticed the test classpath had the wrong version of several dependencies, so I'm guessing that _could_ explain the errors I see there (but not locally). 
There was also complaints about a key store file not being set, which I still think is the most likely root cause, though. I suppose we may have different TLS settings on Factory from what we have locally?

bjorncs (approved):

jonmv:
Also needs https://github.com/vespa-engine/vespa/pull/29434

PR #3139, opened 2023-11-27
Enable gcc toolset by default

baldersheim:
@aressem or @hmusum PR


hmusum (approved):

hmusum:
Fails with:
/bin/sh: /opt/rh/gcc-toolset/enable: No such file or directory

baldersheim:
Expected as it test the PR with latest opensource build which does not have that symlink.

PR #3142, opened 2023-11-28
Test with 100.000 and 1.000.000 values_in_range (posting lists).

geirst:
@baldersheim please review

### DIFF
--- a/tests/performance/range_search/create_docs.cpp
+++ b/tests/performance/range_search/create_docs.cpp
@@ -18,7 +18,7 @@ const IntVector hits_ratios = {1, 2, 4, 5, 6, 8, 10, 20, 40, 50, 60, 80, 100, 15
  * When searching for range(my_field, LOWER, UPPER) this will match 'values_in_range'
  * unique values (or posting lists) and return a number of hits given by the 'hits_ratio'.
  *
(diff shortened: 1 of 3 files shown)


tests/performance/range_search/create_docs.cpp
```diff
@@ -28,7 +28,7 @@ IntVector make_range_values(int num_docs, int values_in_range) {
         int hits = ((size_t)num_docs * (size_t)hits_ratio) / 1000;
         if (hits >= values_in_range) {
             int hits_per_value = hits / values_in_range;
-            int value = hits_ratio * 100000 + values_in_range;
+            int value = hits_ratio * 10000000 + values_in_range;
```
baldersheim, line 31: This will overflow 10M * 10M > 32bit

baldersheim (commented):
Will overflow

tests/performance/range_search/create_docs.cpp
geirst: Max value of `hits_ratio` is 200  and max value of `values_in_range` is 1.000.000. 200 * 10.000.000 + 1.000.000 = 2.001.000.000 (will not overflow).

PR #3181, opened 2024-01-05
- Add method to count how many hits have a matching field_value.

baldersheim:
- Assert that we can at least find ourselves for query/document processing symetry for both indexed and streaming.

@geirst PR

### DIFF
--- a/lib/test_base.rb
+++ b/lib/test_base.rb
@@ -665,7 +665,6 @@ module TestBase
   end
 
   # count number of hits with a matching field value
-  # in a field, for a given hit number.
   def count_hits_with_field_value(query_or_result, fieldname, expectedvalue)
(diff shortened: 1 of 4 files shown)


lib/test_base.rb
```diff
@@ -664,6 +664,17 @@ def assert_field_value(query_or_result, fieldname, expectedvalue, hitnumber=0, e
     assert_equal(to_utf8(expectedvalue), result.hit[hitnumber].field[fieldname].to_s, explanationstring)
   end
 
+  # count number of hits with a matching field value
+  # in a field, for a given hit number.
```
geirst, line 668: "_for a given hit number_" doesn't seem right?

geirst (dismissed):
👍

lib/test_base.rb
baldersheim: No, leftover.

PR #3186, opened 2024-01-11
Revert "Streaming search does not respect case sensitivity yet."

baldersheim:
Reverts vespa-engine/system-test#3185
@geirst Now it does


geirst (approved):
👍

PR #3229, opened 2024-01-17
Revert "Run tensor unstable cell types system test in streaming mode."

toregge:
Reverts vespa-engine/system-test#3217

Times out in streaming mode.

@geirst : please review


geirst (approved):
👍

PR #3252, opened 2024-01-25
Run element completeness system test in streaming mode.

toregge:
@geirst : please review

This PR depends on vespa-engine/vespa#30065


geirst (approved):
👍

PR #3256, opened 2024-01-26
Test operation cancellation on factory

vekterli:
@geirst please review. Will likely have to be reverted within a few hours, but let's try to get a run or two in first.


geirst (approved):
👍

PR #3295, opened 2024-02-07
Avoid using common_merge_chain_optimalization_minimum_size.

baldersheim:
@vekterli PR


vekterli (approved):
:+1: not sure if the test triggers the exact same edge case now that we default to 64, but it's been years since the underlying bug it tests was fixed anyway.

PR #3316, opened 2024-02-10
Run all search tests with streaming to get a full feature overview.

baldersheim:
Reverts vespa-engine/system-test#3309
Give it another go.


hmusum:
👍

PR #3341, opened 2024-02-16
Do not run all tests as streaming

baldersheim:
@hmusum or @geirst PR
Not able to revert #3338, so needed manual intervention


hmusum (commented):
👍

PR #3358, opened 2024-02-22
Disable protobuf document api protocol

hmusum:
Reverts https://github.com/vespa-engine/system-test/pull/3357

Seems like some system tests fail due to this


vekterli (approved):
:+1:

PR #3375, opened 2024-02-27
Remove qrsdocs cleaning, dir has not been used in many years

hmusum:


hmusum:
Hm, seeing one thing I need to check first

hmusum:
Cannot remove, a config has default value that uses this, so need to keep it

aressem:
👍

PR #3423, opened 2024-02-29
Run ann with filter system test only in indexed mode.

toregge:
@geirst : please review


geirst:
This is testing tuning of strategies (pre- vs post-filter), which is an index only feature.

geirst (approved):
👍

PR #3436, opened 2024-03-01
Set test owner

hmusum:
Trying to set owner for all tests, shout if this does not make sense (suggestions to owner are welcome)


geirst (approved):
👍 Suggesting vekterli

PR #3442, opened 2024-03-05
Run interpolated lookup system test partially in streaming mode.

toregge:
@geirst : please review
@baldersheim : FYI

This PR depends on vespa-engine/vespa#30483


geirst (approved):
👍

PR #3456, opened 2024-03-11
Convert feed files to json

hmusum:


tests/docproc/schemamapping/schemamapping.rb
```diff
@@ -14,10 +14,10 @@ def setup
   end
 
   def test_schemamapping_basic
-    feed_and_wait_for_docs("simple", 2, :file => selfdir+"simple.xml", :cluster => "simple", :route => "\"default/chain.apple default/chain.banana indexing\"")
```
baldersheim, line 17: I guess teh xml files should be removed too.

hmusum: Yes, thanks, fixed

PR #3459, opened 2024-03-12
Convert feed files to json, part 4

hmusum:


baldersheim (approved):
We should put put these small files in the code itself and feed directly via ruby, but that is a different step.

PR #3470, opened 2024-03-13
Test using imported fields alongside non-attributes in document selections

vekterli:
@geirst please review. Will fail until https://github.com/vespa-engine/vespa/pull/30611 is in place


geirst (approved):
👍

PR #3538, opened 2024-03-17
Revert "Now streaming and indexed resolves to same cluster name"

baldersheim:
Reverts vespa-engine/system-test#3534
@hmusum or @geirst PR


hmusum:
👍

PR #3607, opened 2024-03-21
Use a well defined field, surl, instead of isbn which is only present…

baldersheim:
… in a subset.

@hmusum or @geirst PR


hmusum:
👍

PR #3757, opened 2024-04-11
Add performance test for feeding with high contention

vekterli:
@geirst please review. Includes a heavily modified version of `create_docs` from `nearest_neighbor_streaming` that allows for shuffling across _all_ docs, not just user IDs.
@baldersheim FYI. Starting out with 500K docs (500 locations * 1K docs).

Test feeding performance when feed has high affinity to specific buckets instead of being evenly distributed. Feeds to a number of locations (ID-specified groups) with many documents in sequence per location. This triggers "worst-case" write lock contention against buckets (amortized 1 bucket per distinct location).

Test is done in 2 phases:
 1. Feed fully shuffled (i.e randomly distributed) to get a baseline.
 2. Re-feed in sequential order.


geirst (approved):
👍

PR #3758, opened 2024-04-11
Use redeploy member function which waits until content nodes have

toregge:
new config generation.

@geirst : please review


geirst (approved):
👍

PR #3778, opened 2024-04-15
Use vespa-feed-client as default feed client

hmusum:
Please review only, want a run on factory to see what breaks


hmusum:
@baldersheim please take a look

baldersheim (approved):

PR #3780, opened 2024-04-16
Use attribute, not index, for numeric fields

hmusum:
Please review only

### DIFF
--- a/tests/search/alias/music.sd
+++ b/tests/search/alias/music.sd
@@ -27,7 +27,7 @@ schema music {
     }
 
     field year type int {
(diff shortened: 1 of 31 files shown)


bratseth:
We should keep "attribute | index" in a few tests though?

baldersheim:
In one test where we also check that a warning is given.

hmusum:
There are 2 tests that keep "attribute | index", but it makes sense to add a test with this and check for warninrg, I'll look into it

hmusum:
@baldersheim please have a look again

baldersheim (approved):

PR #3796, opened 2024-04-18
Reduce load of llama performance temporarily

lesters:
@hmusum Please review. No use in having the performance test like it is currently.

### DIFF
--- a/tests/performance/llama/llama.rb
+++ b/tests/performance/llama/llama.rb
@@ -31,7 +31,7 @@ class LlamaPerformanceTest < PerformanceTest
     deploy_app
     copy_query_file
 
-    warmup
+    # warmup
     run_queries
   end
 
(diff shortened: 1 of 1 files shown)


tests/performance/llama/llama.rb
```diff
@@ -31,7 +31,7 @@ def test_llama_inference
     deploy_app
     copy_query_file
 
-    warmup
+    # warmup
```
hmusum, line 34: No warmup anymore?

hmusum (approved):

tests/performance/llama/llama.rb
lesters: Just removing temporarily, as the warmup also takes like 15 minutes 😂

baldersheim (approved):

tests/performance/llama/llama.rb
hmusum: 👍

PR #3807, opened 2024-04-19
Revert "Convert feed files to json, part 208"

baldersheim:
Reverts vespa-engine/system-test#3804


hmusum (approved):
Thanks, looks like generating weighted set in json does not work

PR #3839, opened 2024-04-26
There must be a better way

jobergum:

### DIFF
--- a/tests/performance/nearest_neighbor_distance_metrics_and_types/nearest_neighbor_distance_metrics.rb
+++ b/tests/performance/nearest_neighbor_distance_metrics_and_types/nearest_neighbor_distance_metrics.rb
@@ -14,20 +14,9 @@ class NearestNeighborDistanceMetricAndTypesPerfTest < PerformanceTest
     deploy_app(SearchApp.new.sd(selfdir + 'vector.sd'))
     start
   
-    feed_file = 'vectors.perf.json.zstd'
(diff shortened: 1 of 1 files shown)


jobergum:
Don't we have a feeder that accepts zstd compressed file? Downloading the 15GB file from S3 to test node takes 9 minutes so it might be that this is too large for perf test runs.

tests/performance/nearest_neighbor_distance_metrics_and_types/nearest_neighbor_distance_metrics.rb
```diff
@@ -13,16 +13,21 @@ def test_metrics_and_types
...
     puts "Running command #{cmd}"
     result = `#{cmd}`
     puts "Result: #{result}"
-    run_feeder(local_file, [])
 
+    cmd2 = "zstdcat '#{local_file}' > '#{local_feed_file}'"
+    puts "Running command #{cmd2}"
+    result2 = `#{cmd2}`
+    puts "Result2: #{result2}"
+    run_feeder(local_feed_file, [])
```
baldersheim, line 30: You can use feed_stream or run_stream_feeder. See
https://github.com/vespa-engine/system-test/blob/master/tests/performance/tensor_update/tensor_update.rb

Something like
feed_stream("curl 'url' | zstdcat")

jobergum:
This looks also much better 👍

PR #3849, opened 2024-04-26
Add dataprep-script for ecommerce benchmarking

thomasht86:

Adding first draft of dataprep-script. 
More details in added README.
It will keep my Mac (and home office) warm over the weekend. 🙂 🔥

FYI @jobergum

### DIFF
--- a/tests/performance/ecommerce_hybrid_search/dataprep/download_and_prep.py
+++ b/tests/performance/ecommerce_hybrid_search/dataprep/download_and_prep.py
@@ -1,0 +1,304 @@
+import os
+import re
+import numpy as np
+import pandas as pd
+from datasets import load_dataset
+from sentence_transformers import SentenceTransformer
+from multiprocessing import Pool
+from typing import Any, Dict, List, Union
+import subprocess
+from dataclasses import dataclass
+import logging
+import argparse
+
+logging.basicConfig(format="%(asctime)s - %(message)s", level=logging.INFO)
+
+# Global configurations
+CACHE_PATH = "./model_cache"
+MODEL_NAME = "Snowflake/snowflake-arctic-embed-xs"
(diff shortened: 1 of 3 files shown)


tests/performance/ecommerce_hybrid_search/dataprep/download_and_prep.py
```diff
@@ -0,0 +1,304 @@
+import os
```
geirst, line 1: Please add copyright header.

tests/performance/ecommerce_hybrid_search/dataprep/download_and_prep.py
```diff
@@ -0,0 +1,304 @@
...
+        df = pd.read_json(file_path, compression="gzip")
+        return ProcessingResult(cat=cat, length=len(df))
+
+    # prices should be a numpy array of 1000 integers between 0 and 1000
+    prices = np.random.randint(1, 1000, 1000)
+
+    def process_entry(entry: Dict[str, Any]) -> Dict[str, Union[str, float, None]]:
+        price_str = entry["price"]
+        try:
+            price = float(re.sub(r"[^\d.]", "", price_str))
```
geirst, line 88: We should multiply the price by 100 to convert it to cents. Can also do it on line 178 when converting to Vespa JSON.

geirst (dismissed):
👍

tests/performance/ecommerce_hybrid_search/dataprep/download_and_prep.py
thomasht86: Conversion from float to int is done in l113. 
Multiplication by 100 was with previous random price generator, but simplified this to just generate ints, so int-conversion will only be applied to the non-imputed prices.

PR #3882, opened 2024-05-03
Revert "Dont report performance data to factory backend when running … MERGEOK

hmusum:
Reverts vespa-engine/system-test#3881

Seems like there is a type (arg should be args), reverting


aressem:
Thanks 👍

PR #3889, opened 2024-05-07
Dataprep updates and ES-skeleton app

thomasht86:

- [x] Updated `download_and_prep.py` - Now does common preprocessing, adds embeddings,  and saves to .parquet. 
- [x] Added `generate_feed_query.py` - This merges all categories, and create vespa/ES feed file format (zstd-compressed) and query files. (ES query files not properly tested yet)
- [x] Also added notebook with skeleton for ES-application. Intention is to take this out to JSON-files and orchestrate with curl or similar from ruby.


tests/performance/ecommerce_hybrid_search/dataprep/generate_feed_query.py
```diff
@@ -0,0 +1,372 @@
...
+        data += json.dumps(doc_data, ensure_ascii=True) + "\n"
+    cctx = zstd.ZstdCompressor()
+    compressed_data = cctx.compress(data.encode("utf-8"))
+    with open(file_name, "wb") as f:
+        f.write(compressed_data)
+    logging.info(f"Data saved in Elasticsearch format to {file_name}")
+
+
+def title_to_query(title: str) -> str:
+    # If the title is more than 8 words, use the first 4 words.
```
geirst, line 131:
```suggestion
    # If the title is more than 8 words, use the first 8 words.

```

tests/performance/ecommerce_hybrid_search/dataprep/generate_feed_query.py
```diff
@@ -0,0 +1,372 @@
...
+
+
+def save_vespa_query_files_from_df(
+    df: pd.DataFrame,
+    query_save_path: Path,
+    num_queries: int,
+    query_type: QueryTypeEnum = QueryTypeEnum.WEAK_AND,
+) -> None:
+    """
+    # 1. weakAnd + bm25 ranking:
```
geirst, line 151: I don't think we need lines 151 to 163, as the same information is found on lines 183 to 200 in the code.

tests/performance/ecommerce_hybrid_search/dataprep/generate_feed_query.py
```diff
@@ -0,0 +1,372 @@
...
+    Format needed by fbench (POST):
+    Ref: https://docs.vespa.ai/en/performance/vespa-benchmarking.html#prepare-queries
+    Two lines per query, example:
+
+    /search/
+    {"query": {"match": {"title": "small money clip leather wallet"}}}
+    /search/
+    {"query": {"match": {"title": "blue nike shoes"}}}
+    """
+    endpoint = "/search/"
```
geirst, line 248: I don't think this is the correct endpoint for ES? The same applies for lines 243 and 245.

tests/performance/ecommerce_hybrid_search/dataprep/generate_feed_query.py
```diff
@@ -0,0 +1,372 @@
...
+    # Validate number of queries requested is less than the total number of samples
+    if num_queries > len(df):
+        raise ValueError(
+            "Number of queries requested is greater than the total number of samples available."
+        )
+    # Sample the dataframe set random state for reproducibility
+    df = df.sample(num_queries, random_state=42)
+    base_parameters = {
+        "size": 10,
+        "fields": ["title", "description"],
```
geirst, line 258: Is this the fields returned in the result? If so, it should only be `"id"`. The same applies for lines 270 and 298.

tests/performance/ecommerce_hybrid_search/dataprep/generate_feed_query.py
```diff
@@ -0,0 +1,372 @@
...
+        "_source": False,
+    }
+    # Generate list of queries
+    queries = df["title"].tolist()
```
geirst, line 262: Should this be `queries = df["title"].apply(title_to_query).tolist()` instead?

geirst (commented):
Overall looks good, some comments to look at first.

tests/performance/ecommerce_hybrid_search/dataprep/generate_feed_query.py
geirst: Or is it the fields being searched? If so, it is correct.

geirst (approved):
👍

PR #3892, opened 2024-05-08
Test setting max-token-length in field match.

toregge:
@geirst : please review

This is related to vespa-engine/vespa#31139


geirst (approved):
👍

PR #3942, opened 2024-05-16
Fix same precision for feed files

thomasht86:

Tested that the new feed file gives same precision, and that it can be fed to Vespa. 
Uploading vespa-1M-file to Drive now.


geirst (approved):
👍

PR #3967, opened 2024-05-20
Revert "Use client params and default_feed_client only"

hmusum:
Reverts vespa-engine/system-test#3966

Please review only, will merge in case this breaks something


baldersheim (approved):

hmusum:
Seems like feeding for some performance tests was impacted, need to investigate

PR #3968, opened 2024-05-20
Switch default feed client to vespa-feed-client

hmusum:
Please review only


baldersheim (approved):
Closing in on target ....

PR #3996, opened 2024-05-23
Use redundancy=2 and searchable-copies=1.

geirst:
This is more similar to ES number_of_replicas=1.

@jobergum please review
@toregge @thomasht86 FYI


baldersheim:
Why redundancy=2 ?

PR #4008, opened 2024-05-24
Remove unused SearchNode softdie and die member functions.

toregge:
@hmusum : please review


baldersheim (approved):

baldersheim:
Thne you can remove the rpc die command from server too.

PR #4033, opened 2024-05-31
Align rank expression in hybrid profile.

geirst:
@thomasht86 please review


thomasht86 (approved):
Good to align these! 👍 
Let's see if rrf works for ES as well.

PR #4040, opened 2024-06-05
Add second phase rank score drop limit system test.

toregge:
@geirst : please review

This is related to vespa-engine/vespa#31454


geirst (approved):
👍

PR #4043, opened 2024-06-05
Use vespa-feeder to stabilize test

hmusum:
@arnej27959 please review

Couldn't find any option for vespa-feed-client that would give same behavior as 'maxpending 1`, @bjorncs or @jonmv might know if that's possible


jonmv (approved):

jonmv:
No, setting a single pending operation isn't possible ... although why shouldn't it be?

bjorncs:
Try to configure client with 1 connection 1 max concurrent stream. That should give a deterministic ordering.

hmusum:
👍 Thanks, done in https://github.com/vespa-engine/system-test/pull/4045

PR #4049, opened 2024-06-07
Revert "Use vespa-feed-client"

aressem:
Reverts vespa-engine/system-test#4048

A few system tests fails and this is the prime suspect.


hmusum:
Yes, it was not deterministic, looking into it

PR #4062, opened 2024-06-14
- Add test showing odd summary behavior when using different summary …

baldersheim:
…classes in grouping and in the flat list.

@geirst PR


baldersheim:
@geirst PTAL, simplified and actually tested, not only logged.

geirst (approved):
👍

PR #4078, opened 2024-06-20
Remove the process cleanup code. We run in ephemeral containers.

aressem:


baldersheim:
I assume this does not prevent test from doing a controlled shutdown. If shutdown times out a stacktrace is taken with pstack and written to the log. It is important that we always try controlled shutdown, and detect failure to do so.

baldersheim:
Also note that developers, at least all c++ developers like myself, @geirst @vekterli @toregge @havardpe @arnej27959 are building and running tests in the same container that is not ephemeral.
I think it would be better that there was an ephemeral flag that factory could pass when running tests.

aressem:
I will find another way around this then.

PR #4094, opened 2024-06-24
Remove visit test that checks lastmodified time

hmusum:
lastmodifiedtime is only available in the deprecated xml output format.


baldersheim (approved):

vekterli (approved):
:+1: ideally we should add this functionality back, but through a different mechanism. It's been effectively void for a long time in visiting due to historical protocol-related reasons

PR #4131, opened 2024-07-12
Revert "Use shared function to download a file from s3."

geirst:
Reverts vespa-engine/system-test#4130
Broke some perf tests.

@vekterli please review
@toregge FYI


vekterli (approved):
:+1:

PR #4132, opened 2024-07-16
Extend struct and map types system test to test weigtedSet and in

toregge:
query operators on struct field (indexed mode only).

This is related to PR vespa-engine/vespa#31974

@vekterli : please review
@geirst : FYI


vekterli (approved):
:+1:

PR #4134, opened 2024-07-16
Also test matched-elements-only with weighted set and in queries

toregge:
on struct fields in streaming mode.

This is related to vespa-engine/vespa#31976

@vekterli : please review
@geirst : FYI


vekterli (approved):
:+1:

PR #4138, opened 2024-07-25
Remove the remote.repositories files to avoid Maven contacting the so…

aressem:
…urces to verify.


freva (approved):
Why not `-delete` option to find instead of `xargs rm -f`?

aressem:
Thanks :) Old habits die hard. Updated to use `-delete`.

freva (approved):

PR #4155, opened 2024-07-30
Don't call show_kernel_core_pattern if active sanitizers

hmusum:
Will just fail for test_coredump_compression, since no node is available


hmusum:
Or maybe we should just remove that line, not sure why it's called in teardown

aressem:
It is probably called there to verify that the kernel core pattern has been reset to what it was previously.

aressem (approved):

PR #4184, opened 2024-08-30
Disable inverted index where not needed

radu-gheorghe:
To match Vespa's attributes without fast-access in an attempt to compare apples to apples.


geirst (approved):
👍

PR #4186, opened 2024-09-03
Change Elasticsearch to dot_product similarity

radu-gheorghe:
With our normalized vectors, the score should be exactly the same as "cosine", but we don't force Elasticsearch to normalize vectors at query time. Which is similar to what our prenormalized-angular does: effectively orders by cosine similarity, assuming vectors are normalized.


geirst (approved):
👍

PR #4187, opened 2024-09-04
Remove all fields from query response to be certain _source is not used.

geirst:
@radu-gheorghe please review


radu-gheorghe (approved):
LGTM :)

PR #4189, opened 2024-09-05
Align nearest neighbor search parameters.

geirst:
This should ensure similar amount of work when searching a HNSW index in Vespa (targetHits=100) and ES (num_candidates=100, k=100).

@radu-gheorghe please review


radu-gheorghe (approved):
Sounds good, that's my understanding, too.

PR #4195, opened 2024-09-10
Test tokens with accents are processed symmetrically

bratseth:


bratseth:
This should break, and be fixed by https://github.com/vespa-engine/vespa/pull/32385, so it would be nice to have this merged first.

PR #4202, opened 2024-09-18
Remove unused env var VESPA_MBUS_DOCUMENTAPI_USE_PROTOBUF

hmusum:
Seems like this is not used anymore


vekterli (approved):
:+1: indeed, this was used prior to finalizing the protocol negotiation for the new protobuf implementation

PR #4208, opened 2024-09-30
Extend test

hmusum:
Test combinations of field, term, sources and query language. Also output result with trace when result is not as expected.


bratseth (approved):
👍

PR #4222, opened 2024-10-10
Relax check for expected hits when refeeding

hmusum:
We cannot know how many hits we will get when refeeding docs, removals and insertions might happen for many docs at the same time, so just check that we have between 0 and the number of hits expected after feeding has finished


vekterli (approved):
:+1:

PR #4225, opened 2024-10-14
Run query filter benchmarks with entire range of clients.

geirst:
@jobergum please review and merge
@radu-gheorghe FYI


jobergum (approved):
🌶️

PR #4230, opened 2024-10-16
Align query benchmarks in ecommerce perf test

geirst:
@radu-gheorghe please review
@jobergum FYI

### DIFF
--- a/tests/performance/ecommerce_hybrid_search/ecommerce_hybrid_search.rb
+++ b/tests/performance/ecommerce_hybrid_search/ecommerce_hybrid_search.rb
@@ -27,7 +27,7 @@ class EcommerceHybridSearchTest < EcommerceHybridSearchTestBase
 
     feed_thread = Thread.new { benchmark_feed(feed_file_name, "refeed_with_queries") }
     sleep 2
-    benchmark_queries("during_refeed", false, [1, 16, 64], {:runtime => 9})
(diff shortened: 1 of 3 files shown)


tests/performance/ecommerce_hybrid_search/ecommerce_hybrid_search.rb
```diff
@@ -15,8 +15,6 @@ def test_hybrid_search
     dump_thread_stats
 
     benchmark_feed(feed_file_name, "feed")
-    benchmark_queries("after_feed", false, [1])
```
radu-gheorghe, line 18: Doesn't this mean that we only run un-filtered queries after flush? I think we want both filtered and unfiltered after flush? Or just filtered?

radu-gheorghe (approved):
I have a small comment, not sure if it's my lack of understanding or there's indeed something to change. Otherwise it looks good, thanks!

tests/performance/ecommerce_hybrid_search/ecommerce_hybrid_search.rb
geirst: All the relevant query benchmarks (both filtered and un-filtered) are found on lines 22 and 23. Lines 18 and 19 are just remains of benchmarking done before flushing, which is not relevant. I remove them to reduce the runtime of the test.

PR #4257, opened 2024-11-05
remove score limit

havardpe:
also adjust X variant to combine values tested separately

@arnej27959 please review


arnej27959 (approved):
👍

PR #4272, opened 2024-11-14
Reduce zookeeper-pre-alloc-size from 64k to 16k

hmusum:
Please review only, will try this out when factory is stable


vekterli (approved):
:+1:

PR #4274, opened 2024-11-15
Benchmark queries that hit on-disk bit vectors

vekterli:
@geirst please review.
@toregge FYI

Bit vector cache is set up with 10% of posting list cache size.

Adds and uses a non-BM25 rank profile. To cut down on testing time, this also removes testing the `NORMAL` IO mode, as we really only consider Mmap or explicit caching as our use-cases.

### DIFF
--- a/tests/performance/mmap_vs_directio/mmap_vs_directio.rb
+++ b/tests/performance/mmap_vs_directio/mmap_vs_directio.rb
@@ -94,11 +94,11 @@ class MmapVsDirectIoTest < PerformanceTest
     puts "Searching with '#{pretty_mode}' search store backing using #{clients} clients"
     puts "Search using BM25 rank profile (full queries)"
     report_io_stat_deltas do
-      benchmark_queries(@query_file_name, run_type, clients, false, @profile[:query_runtime], 'minimal')
+      benchmark_queries(@query_file_name, run_type, clients, false, @profile[:query_runtime])
     end
     puts "Search using BM25 rank profile (queries without stop-words)"
     report_io_stat_deltas do
-      benchmark_queries(@no_stop_words_query_file_name, "#{run_type}_no_stop_words", clients, false, @profile[:query_runtime], 'minimal')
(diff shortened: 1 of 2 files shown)


tests/performance/mmap_vs_directio/mmap_vs_directio.rb
```diff
@@ -157,7 +165,7 @@ def benchmark_queries(query_file, type, clients, warmup = false, runtime = 20)
     run_fbench2(@container,
                 node_file,
                 {:clients => clients,
-                 :append_str => '&presentation.summary=minimal&hits=10',
+                 :append_str => "&presentation.summary=#{summary}&hits=10",
```
geirst, line 168: `presentation.summary` should always be `minimal` for this test. However, we need an extra parameter: `&ranking.profile` which is `default` by default and `use_bitvectors` in the filter test.

tests/performance/mmap_vs_directio/mmap_vs_directio.rb
```diff
@@ -94,11 +92,17 @@ def deploy_and_run_queries(search_io_mode:, cache_size_mb: 0)
...
 
     puts "Searching with '#{pretty_mode}' search store backing using #{clients} clients"
+    puts "Search using BM25 rank profile (full queries)"
+    report_io_stat_deltas do
+      benchmark_queries(@query_file_name, run_type, clients, false, @profile[:query_runtime], 'minimal')
+    end
+    puts "Search using BM25 rank profile (queries without stop-words)"
     report_io_stat_deltas do
-      benchmark_queries(@query_file_name, run_type, clients, false, @profile[:query_runtime])
+      benchmark_queries(@no_stop_words_query_file_name, "#{run_type}_no_stop_words", clients, false, @profile[:query_runtime], 'minimal')
```
geirst, line 101: Either remove the last parameter or change it to `default`.

geirst (approved):

PR #4279, opened 2024-11-18
Handle no directio padding in stats for cached disk io.

toregge:
@geirst : please review

This tracks changes introduced in vespa-engine/vespa#32877


hmusum (approved):
👍

PR #4301, opened 2024-12-09
Use unranked rank profile.

toregge:
@geirst : please review


toregge:
The test uses rank cutoff in the default rank profile, thus the need to use the unranked rank profile after feeding to check if all documents are present.

PR #4342, opened 2025-02-12
Update expected query error message

vekterli:
@bratseth please review. Reflects changes in https://github.com/vespa-engine/vespa/pull/33300


bratseth (approved):
Thanks!

PR #4343, opened 2025-02-13
Remove feature flag setting for container-dump-heap-on-shutdown-timeout

hmusum:
Use same as default value (false)

### DIFF
--- a/docker/include/feature-flags.json
+++ b/docker/include/feature-flags.json
@@ -29,7 +29,6 @@
         { "id" : "rpc-events-before-wakeup", "rules" : [ { "value" : 1 } ] },
         { "id" : "async-message-handling-on-schedule", "rules" : [ { "value" : true } ] },
         { "id" : "max-uncommitted-memory", "rules" : [ { "value" : 130000 } ] },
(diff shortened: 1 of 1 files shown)


bjorncs (approved):

docker/include/feature-flags.json
```diff
@@ -29,7 +29,6 @@
         { "id" : "rpc-events-before-wakeup", "rules" : [ { "value" : 1 } ] },
         { "id" : "async-message-handling-on-schedule", "rules" : [ { "value" : true } ] },
         { "id" : "max-uncommitted-memory", "rules" : [ { "value" : 130000 } ] },
-        { "id" : "container-dump-heap-on-shutdown-timeout", "rules" : [ { "value" : true } ] },
```
bjorncs, line 32: We're disabling the feature?

hmusum: Yes, to use the same in tests as in production. Unless it makes sense to do this on factory to actually generate heap dumps when we get timeout on shutdown. Does it?

bjorncs: Hmm, yeah you're right, I can't remember last time we processed a heap dump from factory.

PR #4354, opened 2025-02-20
Revert "Change summary-decode-policy feature flag to 'eager' (which is the de…"

hmusum:
Reverts vespa-engine/system-test#4347

Go back to on-demand, which is what we will try moving towards as the new default.

@geirst please review


hmusum:
@geirst please review

geirst (approved):

PR #4365, opened 2025-02-28
Track state explorer changes for hnsw index.

toregge:
@vekterli : please review
@geirst : FYI

This PR should be merged at the same time as vespa-engine/vespa#33450


vekterli (approved):
:+1:

PR #4367, opened 2025-03-05
Increase timeout

hmusum:
Test failed due to timeout, log says that restart took more than 30 seconds, so increase it.


arnej27959 (approved):
👍

PR #4368, opened 2025-03-05
Arnej/use opennlp stemming

arnej27959:
@hmusum please review
@bratseth FYI
after verifying what is going on I think it makes most sense to just accept current behavior.


hmusum (approved):
Nice!

PR #4371, opened 2025-03-09
Test binarization

bratseth:
Test implicit and explicit binarization. Merging to get a factory run.


arnej27959 (approved):
Looks good to me

PR #4381, opened 2025-03-23
Test cased index tokens

bratseth:
Should work with https://github.com/vespa-engine/vespa/pull/33650

### DIFF
--- a/tests/search/indexed_tokens/indexed_tokens.rb
+++ b/tests/search/indexed_tokens/indexed_tokens.rb
@@ -10,6 +10,7 @@ class IndexedTokensTest < IndexedStreamingSearchTest
   def feed_doc(idsuffix, doc_template)
     doc = Document.new("test", "id:test:test::#{idsuffix}").
       add_field("stext", doc_template[:stext]).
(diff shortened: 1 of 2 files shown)


tests/search/indexed_tokens/indexed_tokens.rb
```diff
@@ -35,13 +36,15 @@ def test_lingustics_tokens
                          component(Component.new("com.yahoo.language.opennlp.OpenNlpLinguistics"))))
     start
     feed_doc("0", { :stext => "Hello world",
+                    :stext => "Hello cASed",
```
toregge, line 39: Duplicate hash key. This will cause the test to fail when checking the `stext_tokens` summary field later on.

toregge (changes requested):

tests/search/indexed_tokens/indexed_tokens.rb
bratseth: Thanks!

bratseth:
@tegge PTAL

bratseth:
@toregge PTAL

toregge (approved):

PR #4388, opened 2025-04-03
Add json select query test for IN operator

boeker:
Adds a json select query test that basically just tests the IN operator. Should be extended with more examples. Uses a custom schema since the IN operator does not work for string fields with `match:text`.


hmusum (approved):
Looks great! Easily readable and nicely structured

PR #4389, opened 2025-04-03
Add timeout (streaming queries need higher timeout)

hmusum:


boeker (approved):
I got the streaming test to fail locally with a very small timeout while the indexed test still was working, so this really seems to be the culprit. 👍

hmusum:
Nice, thank! Merging

PR #4395, opened 2025-04-07
Fix path [MERGEOK]

hmusum:
Fixes broken build (after changes in https://github.com/vespa-engine/system-test/pull/4394)


glebashnik (approved):
Thanks for the fix

PR #4396, opened 2025-04-07
Add test case for reindexing of new, trivially true bool field

boeker:
The test case checks that, if a trivially true bool field is added to a schema, then the  correct value (`true`) is assigned to old documents after reindexing.

### DIFF
--- a/tests/search/reindexing/reindexing.rb
+++ b/tests/search/reindexing/reindexing.rb
@@ -13,6 +13,51 @@ class ReindexingTest < IndexedStreamingSearchTest
     set_owner('bjorncs')
   end
 
+  def test_add_const_field
+    set_description("test that adds a constant bool field to documents and verifies that the correct value is assigned during reindexing")
+    testdir = selfdir + "add_const_field/"
+
+    # First, we use the schema without the field const_bool
+    system("cp #{testdir}item.0.sd #{dirs.tmpdir}item.sd")
+    app = SearchApp.new.sd(dirs.tmpdir + "item.sd")
+    deploy_app(app)
+
+    start
+
+    puts "Feeding single document"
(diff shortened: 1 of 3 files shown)


tests/search/reindexing/reindexing.rb
```diff
@@ -13,6 +13,51 @@ def setup
     set_owner('bjorncs')
   end
 
+  def test_add_const_field
+    set_description("test that adds a constant bool field to documents and verifies that the correct value is assigned during reindexing")
+    testdir = selfdir + "add_const_field/"
+
+    # First, we use the schema without the field const_bool
+    system("cp #{testdir}item.0.sd #{dirs.tmpdir}item.sd")
```
hmusum, line 21: This is one of the non-intuitive things about the system test framework. Running the test happens on the a `controller` node, while Vespa is running another node. Commands are run with RPC. Copying files needs to be done with something like:

```suggestion
    vespa.nodeproxies.first[1].copy_remote_file_to_local_file(selfdir + "/add_const_field/item.0.sd", dirs.tmpdir + "item.sd")
```

the same goes for the other line using `cp`

boeker: I took this part from [schemachanges_struct_attributes.rb](https://github.com/vespa-engine/system-test/blob/master/tests/search/schemachanges/schemachanges_struct_attributes.rb) (and [schemachanges_base.rb](https://github.com/vespa-engine/system-test/blob/master/tests/search/schemachanges/schemachanges_base.rb)). `vespa.nodeproxies.first[1].copy_remote_file_to_local_file` isn't used there, is it? Why do these tests work?

hmusum (approved):

PR #4402, opened 2025-04-14
Rename dispatch test

hmusum:
We had two dispatchers for some time, now we only have one, so rename to make it easier to find


boeker (approved):
👍

PR #4406, opened 2025-04-23
Bjorncs/grouping filter

bjorncs:


havardpe (approved):
could we use check_grouping instead of check_query more places?

bjorncs:
> could we use check_grouping instead of check_query more places?

We can consider moving `check_grouping` to the test library for easier reuse in the future. It's currently only part of the struct_and_map_types test.

PR #4410, opened 2025-04-24
Rewrite test to check json properly and add back ignored tests

hmusum:
Need to figure out what syntax should be for removing an element in an array with structs, doc is missing/unclear, but I think this could be merged now


arnej27959 (approved):
looks good!

PR #4414, opened 2025-04-24
Don't check state v1 resources that is not /state/v1/custom

hmusum:
This is handled by new system test in tests/search/statev1/statev1.rb now


boeker (approved):
👍 , merging

PR #4415, opened 2025-04-25
Stop using gatewayxmlparser, part 3

hmusum:
Had to rewrite some code due to how handling of removes in response to visit commands was done, probably more can be done, but tried to do as few changes as possible


arnej27959 (approved):
👍

PR #4421, opened 2025-04-27
Add test for max document size

hmusum:


tests/search/max_document_size/max_document_size.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    # feed a bit less than 1 MiB
+    feed(@one_mib - 100)
+
+    # feed a bit more than 1 MiB, should fail
+    begin
+      feed(@one_mib + 100)
+      raise "An error should have been raised when feeding more than 1 MiB"
+    rescue HttpResponseError => e
+      # as expected
```
vekterli, line 26: Should we test the concrete HTTP response code generated?

vekterli (approved):
:+1:

tests/search/max_document_size/max_document_size.rb
hmusum: Yes, that as well as verify the error message. Right now feed of doc > 1MiB does NOT fail, so need to figure out what's wrong with test or config/code.

hmusum: Added

hmusum:
Test now works as expected, except getting http status code 500 instead of 400 (known issue).

PR #4423, opened 2025-04-28
Update expected output from feed with division by zero

hmusum:
"Division by zero" -> "arithmetic divide 0.0"

Maybe the output should not have changed, in that case close this PR and fix the code instead


bratseth (approved):
👍

PR #4428, opened 2025-04-30
Set element gap for index environment.

toregge:
This should ber merged at the same time as vespa-engine/vespa#33913


vekterli (approved):
:+1:

PR #4430, opened 2025-05-02
Test element gap for near search.

toregge:
@vekterli: please review

This PR depends on vespa-engine/vespa#33927


vekterli (approved):
:+1:

PR #4431, opened 2025-05-08
Track update to summary config: separate element filtering config.

toregge:
@havardpe : please review
@arnej27959 : FYI

This tracks the changes in vespa-engine/vespa#33970


arnej27959 (approved):
👍

PR #4435, opened 2025-05-15
fix test so it actually fails when the Searcher says "bad"

arnej27959:
this should make the test fail for streaming as expected...
@vekterli please review
@bratseth FYI


vekterli (approved):
:+1:

PR #4441, opened 2025-05-23
Remove sleep

hmusum:
Remove sleep, as assert_numdocs loops until expected number of docs are found. Also reduce sleep time in loop


vekterli (approved):
:+1:

PR #4463, opened 2025-06-18
Add temporary `perf stat` of CPU power delivery levels for grouping test

vekterli:
@arnej27959 please review. Note: I have not been able to test this on a bare-metal node (no usable low-level PMUs on VMs in The Cloud™️...), so this may or may not work 👀 Have cross-checked the event/umasks with https://perfmon-events.intel.com/ for Ice Lake, which is what it appears the perf tests currently run on.

A wild stab in the dark to figure out mysterious transitive(?) performance behavior of some new AVX-512 code that is barely even touched by the test.

Measure cycles run when the CPU is in a non-baseline (power license level > 0) power state, which may happen during certain AVX processing situations. But really shouldn't happen here...


arnej27959 (approved):
👍

PR #4464, opened 2025-06-19
Move onnx model test from internal repo

hmusum:
Rewritten,  based on `tests/search/global-phase-ranking/global-phase-ranking.rb`, but simplified a lot


hmusum:
@glebashnik please take a look again

PR #4467, opened 2025-06-20
Add flat `perf record` output around `fbench` in test

vekterli:
@arnej27959 please review. Adds `perf record/report` output and also adds level 0 power license cycle counting to have more numbers to compare.

Xeon power license PMUs are still tracked, but only for the first 15 seconds. Also predicate PMU `perf stat` on the underlying CPU being x86-64. Just skip it entirely otherwise.


arnej27959 (approved):
👍

PR #4490, opened 2025-06-28
report results to factory even for perf recording runs

arnej27959:
Also:
- Fix typo in variable name
- Avoid duplicate perf report at end of test
- Some more logging
- Add to filtering of too-long lines in perf report

@toregge or @hmusum please review


hmusum (approved):
👍

PR #4498, opened 2025-06-30
Arnej/avoid multithreade perf recording

arnej27959:
@glebashnik please review


glebashnik:
@arnej27959 Do we need to add `if` for `write_report(fillers)` in `es_feed_helper`?

PR #4516, opened 2025-07-16
Add test of vespa-print-default

hmusum:


tests/config/tools/vespa_print_default.rb
```diff
@@ -0,0 +1,46 @@
...
+
+class VespaPrintDefault < ConfigTest
+
+  def setup
+    set_owner("musum")
+    set_description("Tests vespa-print-default")
+  end
+
+  def test_vespa_config
+    app_gen = SearchApp.new.sd(SEARCH_DATA+"music.sd")
```
toregge, line 14: no need to deploy an application, e.g. https://github.com/vespa-engine/system-test/blob/0db2a63116ac294d1a1aced22423cdbab0d0bd89/tests/search/config_definitions_not_changing/config_definitions_not_changing.rb#L12

hmusum: I am unable to get the correct hostname (will get just `localhost`) if I use what you suggest, I tried it, but reverted back to deploying an app. Maybe there is some other way...

toregge: That's just a matter of doing about the same as vesa-model does at
https://github.com/vespa-engine/system-test/blob/0db2a63116ac294d1a1aced22423cdbab0d0bd89/lib/vespa_model.rb#L787
i.e. call `set_addr_configservers` on the node proxy.

PR #4527, opened 2025-08-11
Add system test to check that nearestNeighbor does not return documents with empty field

boeker:
This test makes sure that the nearestNeighbor operator does not return documents where the searched tensor field is empty. Streaming search, exact search, and approximate search are tested.

Depends on https://github.com/vespa-engine/vespa/pull/34574.


arnej27959 (approved):
Looks good to me

PR #4531, opened 2025-08-14
streaming does not insert NaN for missing numbers

arnej27959:
@bjorncs please review


tests/search/grouping_adv/grouping_indexed.rb
```diff
@@ -27,6 +27,14 @@ def test_advgrouping_fs4
 
     querytest_common
 
+    # Test buckets with map value
+    check_query('all(group(predefined(msd{k1},bucket(1,3),bucket(6,9))) each(output(count())))', 'indexed-predef5')
```
bjorncs, line 31: Consider making the last of `check_query` argument a file name suffix without the 'streaming-'/'indexed-' prefix - `predef5` in this case. When executing the query against indexed document type, use answer file `indexed-predef5.json` if exists, otherwise fall back to `predef5.json`. Do same for streaming with preference for `streaming-predef5.json`.

The benefit is that we avoid duplicate queries in `grouping_indexed.rb` and `grouping_streaming.rb`.

bjorncs (approved):

PR #4558, opened 2025-09-09
Remove test variant using lz4

hmusum:
Same performance characteristic as test without lz4, has little value as a performance test.


arnej27959 (approved):
👍

PR #4561, opened 2025-09-09
Reduce number of values to test for range_search and in_operator tests

hmusum:
Similar performance characteristics for these values, seems unnecessary to use time on this,
this will also reduce total runtime for tests.


arnej27959 (approved):
👍

PR #4589, opened 2025-09-19
Revert "Use mimalloc in system tests"

hmusum:
Reverts vespa-engine/system-test#4588

Looks like this caused some core dumps


vekterli (approved):
:+1: glibc malloc seems to have snuck in

hmusum:
Will look into this, reverting for now

PR #4616, opened 2025-10-07
Use custom labels in ANN removal test

boeker:
Add custom labels for before and after removal so that the data points can be distinguished. Use a hash for all optional parameters to also make the tests more readable.

@arnej27959 please review and merge


arnej27959 (approved):
👍

PR #4851, opened 2026-03-25
Fix recall computation in NearestNeighborRecallSearcher

boeker:
The current implementation in NearestNeighborRecallSearcher reports a too-low recall if multiple hits have the same relevance. This already happened in the sift test in my testing, but the difference is not that large (roughly 92.7 before vs 93.7 with the change for 100 target hits and no filtering). This becomes more important when using it to compute the recall for embeddings where a lot of results have the same distance.

An alternative could be to sort the results by relevance first and, if the relevance is the same, by the id, but this approach seems simpler.


arnej27959 (approved):
agreed 👍

PR #4857, opened 2026-04-09
Clean up and "ruby-ify" hierarchical distribution system test

vekterli:
@hmusum please review. Should be no functional changes.

Remove redundant cluster sync waits that should be subsumed by other sync operations.


hmusum (approved):
👍

PR #4859, opened 2026-04-15
Wait for hitcount after removal in ANN removal test

boeker:
Tests still fails every now and then. Let's try to wait for the hitcount after removing the documents instead of asserting it.


hmusum (approved):
👍

PR #4866, opened 2026-04-19
Use regexp as first argument to asserts

hmusum:
String works now, but need to change before planned removal of lib/assertions.rb


johansolbakken (approved):
👍🏻

PR #4869, opened 2026-04-20
Add system test of pseudo row-column mode for down/retired node states

vekterli:
@toregge please review

This tests that fundamental properties of the ideal state algorithm also hold in pseudo row-column mode for nodes in down/retired states.

For both states, other nodes within a group (row) must take over buckets. For down state, the minimal data movement properties of the algorithm shall hold. This is however not the case for retired states.

Also ensure that we run hierarchical distribution tests with a distribution bit level that is high enough that all fed documents end up in their own logical bucket (this seems to have been lost when tests were moved to 8 distribution bits many moons ago to speed up visiting and maintenance ops). Doesn't affect correctness, but for all things pseudo-random, more sample points == better...!


toregge (approved):
👍

PR #4871, opened 2026-04-21
Test that maintenance in pseudo row-column does not move buckets

vekterli:
@toregge please review. Basically just checks that a whole lot of nothing happens when a node is set into maintenance.


toregge (approved):
👍

PR #4874, opened 2026-04-22
Start services after deploying

hmusum:
Try to avoid error messages in log due to app not being deployed yet


johansolbakken (approved):

arnej27959:
You need to deploy without wait first, then start, then maybe you can deploy again with waiting

PR #4895, opened 2026-05-04
Fix assert() (use assert_equal)

hmusum:
Fixed to use assert_equal, but now test fails with:

  1) ERROR IN 'Redundancy::test_kill_node__PROTON()': AssertionFailedError: <0> expected but was
<2>.

     at /yahoo-systemtests/system-test/tests/vds/redundancy.rb:124:in `test_kill_node'

Help! 🙃

### DIFF
--- a/tests/vds/redundancy.rb
+++ b/tests/vds/redundancy.rb
@@ -119,9 +119,11 @@ class Redundancy < PersistentProviderTest
     assert(statinfo.has_key?("0"))
     assert(statinfo.has_key?("1"))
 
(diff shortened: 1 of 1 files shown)


vekterli (dismissed):
:+1:

hmusum:
@vekterli please take a look, update assert and check that status for doc is DELETED

vekterli (approved):
:+1: turns out the `DELETED` string is emitted by the system test framework for tombstones in `stat`; had to double-check since I couldn't remember the backend emitting this value

PR #4899, opened 2026-05-05
Add system test for first phase max

johansolbakken:
Use `firstPhaseMax` in the second phase to use the highest first-phase rank score in rank expressions.

Added `test_dynamic_relative_threshold`, which is the use-case related to the feature request.


hmusum (approved):
👍

PR #4903, opened 2026-05-08
Make metricsproxy service count assertion use a minimum bound

arnej27959:
In factory deployments, configserver does not run locally on the test node, so the exact expected count of 12 services does not hold. Replace the exact-equality assertions with a minimum bound (>= 9) so the test passes in both local and factory environments without hard-coding every possible variant.

Also add diagnostic puts-logging of which services were found, to make future failures easier to diagnose, and answer the long-standing TODO comment: the one-service discrepancy between the metrics proxy port and the container port is caused by an extra "host_life" virtual service.


hmusum (approved):
Great, thanks for adding the updated comment

PR #4904, opened 2026-05-11
Johsol/grouping function node empty arg

johansolbakken:
*🤖 AI declaration: this code is co-written with claude code*

@arnej27959 please review


arnej27959 (dismissed):
this looks good but consider waiting until we have clarified if current behavior is the one we want..

johansolbakken:
@arnej27959 updated expected to match the implementation.

arnej27959 (approved):
👍

PR #4906, opened 2026-05-12
Revise test for document-id attribute

boeker:
Uses the in-schema setting now. Adds some test cases. Cleans up the test.


tests/search/documentid_attribute/documentid_attribute.rb
```diff
@@ -0,0 +1,188 @@
...
+    @schema_dir = selfdir + "schemas/"
+    @doc = Document.new("id:storage_test:test:n=1234:1").add_field("some_field", 42)
+    @dms_filename = "[documentmetastore].dat"
+    @dms_docid_filename = "[documentmetastore].docids.dat"
+    @dms_docid_file_size = 0x1000 + 4 + 29 # Header + bytes for length of string + bytes for actual string
+  end
+
+  def deploy_with(document_id_setting)
+    puts "# Deploying with document-id setting '#{document_id_setting}'"
+    system("cp #{@schema_dir}test.#{document_id_setting}.sd #{dirs.tmpdir}test.sd")
```
toregge, line 18: cp would not be needed if the sd files had the same base name, but were located in different directories instead.

toregge (approved):

PR #4908, opened 2026-05-12
update expected values from (str_int_map.value + int_array)

arnej27959:
Note that this specific expression isn't really meaningful anyway, since the order of the array you get from str_int_map.value is undefined. However, the test reveals that the values we get from the array with a bound index are returned embedded in a 1-sized array, when possibly it would be better if they became scalar values.


hmusum (approved):

johansolbakken (approved):
👍🏻

PR #4912, opened 2026-05-14
Improve class mismatch error in stat.rb to include path and values

hmusum:
## Summary

- Add a `path` parameter threaded through `entry_delta`, `map_delta`, and `array_delta` so the error pinpoints exactly which key or array index has a type mismatch
- Error message now includes the full path (e.g. `cpu -> user -> [3]`) and the `inspect` values of both sides
- Switch `array_delta` to `each_with_index` on the zipped result to make the index available for path tracking

*This PR was generated with Claude Code*

### DIFF
--- a/lib/performance/stat.rb
+++ b/lib/performance/stat.rb
@@ -254,31 +254,33 @@ module Perf
         @fields[x]
       end
 
-      def entry_delta(this, that)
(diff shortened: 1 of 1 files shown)


johansolbakken (approved):
👍🏻

PR #4916, opened 2026-05-20
Add warning about SVE vector length to list of ignorable log messages

boeker:
This warning can pop up on AlmaLinux 9 (when virtualized on a modern Mac?).


vekterli:
IIRC the Apple Hypervisor on M4 erroneously reports SVE2 (but not SVE) as part of its virtualized AArch64 feature registers to the guest kernel, causing some confusion. No M-series CPUs have SVE/SVE2, though >=4 does have SME. Could double-check what `/proc/cpuinfo` reports on the guest.

arnej27959 (approved):

PR #4921, opened 2026-05-26
Add system test for timing-related ANN parameters

boeker:
Depends on https://github.com/vespa-engine/vespa/pull/36993.

Became a bit overkill this test. Let's see how stable it will be. The current epsilon values are quite generous.


havardpe (approved):
forcing stuff to time out can be both verbose and brittle. If we want to observe the effects of a time out, we could consider a hidden parameter that forces the operation to time out artificially in order to observe the fall-out.

PR #4923, opened 2026-05-28
Update ANN timing test

boeker:
Verifies that timed out query is reported as degraded. Also increases the epsilon.


boeker:
@havardpe Updated the test to run each query multiple times. Should be pretty stable now. :)

havardpe (approved):

PR #4927, opened 2026-06-01
Rescue error when getting procfs report and dump values

hmusum:
We sometimes cannot get some or all values from procfs, write them out instead of failing

### DIFF
--- a/lib/performance_test.rb
+++ b/lib/performance_test.rb
@@ -299,7 +299,7 @@ class PerformanceTest < TestCase
       puts '--------'
       puts delta.printable_result
     rescue RuntimeError => e
-      puts 'Unable to create system report'
+      puts "Unable to create system report: #{e.message}"
       puts "Snapshot at setup time: #{@node_procfs_snapshot_at_setup_time}"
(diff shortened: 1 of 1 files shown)


lib/performance_test.rb
```diff
@@ -293,10 +293,16 @@ def setup
...
-    puts 'System report for duration of test case on test runner node:'
-    puts '--------'
-    puts delta.printable_result
+    begin
+      delta = Perf::Stat::snapshot_period(@node_procfs_snapshot_at_setup_time, procfs_snapshot_now)
+      puts 'System report for duration of test case on test runner node:'
+      puts '--------'
+      puts delta.printable_result
+    rescue RuntimeError => e
+      puts 'Unable to create system report'
```
vekterli: Should we include the exception message here for debugging?

vekterli (dismissed):
:+1:

lib/performance_test.rb
hmusum: Yes, good point, adding it

vekterli (approved):

PR #4930, opened 2026-06-04
Fix assert (use assert_equal)

hmusum:
Use assert_equal and update to take into account tombstoned docs (basically same change as in https://github.com/vespa-engine/system-test/pull/4895)


vekterli (approved):
:+1:

PR #4935, opened 2026-06-09
Repair ANN with filter test

boeker:
Was broken by changed event messages.

Adapts the test to the new messages and sets owner to me.


hmusum (approved):
👍

PR #4960, opened 2026-06-25
Add map search performance test

arnej27959:
Add a new performance test (tests/performance/map_search) that measures the cost of searching map<string,int> attribute fields in three different ways: by value, by sameElement(key, value), and by key AND-combined with an independent selective filter term.


johansolbakken (approved):
This looks great 👍🏻

PR #4964, opened 2026-06-29
Add CPU diagnostics to debug bi-modal TensorMatrixMatrixProduct results

arnej27959:
The TensorMatrixMatrixProduct performance test sometimes "flips" between two performance levels between runs. Gather more system/CPU state to find the cause:

- performance_test.rb: log "numactl -s" for each node during setup, next to the existing os-release/uname/lscpu output, to capture CPU/memory pinning and NUMA policy.
- tensor_matrix_matrix_prod: run "turbostat sleep 10" in a loop on a background thread while warmup and queries execute, to observe CPU frequency / turbo-boost behaviour during the benchmark. The thread is killed once run_queries returns.

This is diagnostic instrumentation only; it does not change what the test measures.

### DIFF
--- a/lib/performance_test.rb
+++ b/lib/performance_test.rb
@@ -286,6 +286,7 @@ class PerformanceTest < TestCase
     remote_node = vespa.nodeproxies.values[0]
     remote_node.execute("grep '^PRETTY_NAME' /etc/os-release")
     remote_node.execute('uname -a')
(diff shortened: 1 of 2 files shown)


hmusum (approved):

vekterli (approved):
:+1:

PR #4968, opened 2026-06-30
Sample CPU turbo-license levels during fbench run [MERGEOK]

arnej27959:
While running the TensorMatrixMatrixProduct performance test we suspect the heavy AVX-512 kernels in the matrix-product rank profiles may be triggering the CPU to drop into a lower turbo ("power license") frequency band, which would skew the measured throughput.

To investigate, spawn a background thread that runs "perf stat" against the proton process and counts the core_power_lvl0/1/2_turbo_license PMU events for ~21s, overlapping the matrix_product_512_float_outer_outer fbench run. The counts are dumped to the test output; all this only for x86-64 hosts.

Also: dump turbostat output less often, to avoid overflowing the log.


vekterli (approved):
:+1:

PR #4975, opened 2026-07-07
map_search perf test: add large-range query types (256 and 512 values)

arnej27959:
The existing :range and :combined_range query types only span 21 values, so they say little about how performance scales with range width. Add :large_range and :combined_large_range, which match intervals of 256 or 512 values, to measure that:

- :large_range uses sameElement with range() on the plain map field.
- :combined_large_range matches the same documents via the combined "key#%08x" string attribute using prefix matching instead of the "in" operator (which would need 256/512 terms). This works because the intervals are aligned to hex-digit boundaries: 768-1023 is "...000003xx" and 1024-1279 is "...000004xx".

Also trim the value sweep for the existing query types from [500, 700, 900, 1000] to [700, 1000] to keep total run time in check now that more query types are run (500, at 5 stddev below the value mean, was always skipped for too few expected hits anyway).


johansolbakken (approved):
👍🏻

PR #4979, opened 2026-07-14
Fix crash when subtracting metrics with drifting array length

hmusum:
## Summary
- `ProgrammaticFeedClientTest::test_throughput` (and any performance test using `Perf::System`) intermittently crashed with `Class mismatch when subtracting metrics at cpu -> intr -> [301]: Integer (21) != NilClass (nil)`.
- Root cause: `/proc/stat`'s `intr` (per-IRQ) array length can drift between two snapshots as the kernel adds/removes IRQ vectors at runtime. `array_delta` used `Array#zip`, which pads a shorter array with `nil`, and `entry_delta` treated `Integer` vs `NilClass` as a fatal type mismatch — even though this field isn't used by any reported metric.
- Fix: `entry_delta` now skips an entry when either side is missing (mirroring the existing behavior for hash keys in `map_delta`), and `array_delta` compacts those skipped entries out. Genuine type mismatches (e.g. `Hash` vs `Array`) still raise.
- Added `lib/performance/test/stat_test.rb`, verified it reproduces the original crash message when run against the pre-fix code and passes with the fix.

Generated by Claude.

### DIFF
--- a/lib/performance/stat.rb
+++ b/lib/performance/stat.rb
@@ -255,6 +255,13 @@ module Perf
       end
 
       def entry_delta(this, that, path = [])
+        # One side missing (not both) happens when array_delta zips two arrays of
+        # different lengths (Array#zip pads the shorter side with nil) -- e.g.
+        # /proc/stat's 'intr'/'softirq' per-IRQ arrays, whose length can drift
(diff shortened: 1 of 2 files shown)


boeker (approved):
👍

PR #4992, opened 2026-08-18
Update expected results

hmusum:
Update after changes in https://github.com/vespa-engine/vespa/pull/37481

Change reverted in https://github.com/vespa-engine/vespa/pull/37607, so might do that instead


hmusum:
Obsoleted by https://github.com/vespa-engine/system-test/pull/4994

PR #4994, opened 2026-08-19
Rewrite to use and compare json responses

hmusum:
After rewriting to json this test works both with and without the changes in https://github.com/vespa-engine/vespa/pull/37481. I guess the json renderer handles numbers differently (?)


toregge (approved):
👍

PR #4995, opened 2026-08-20
Relax embedding value tolerance report all differences

glebashnik:

Changed tolerance to 1e-4.
Also compares and reports differences in all values instead of the first mismatch.
Previous tolerance 1e-5 is a bit too tight, catching noise from onnx runtime upgraded.


arnej27959 (approved):
👍

PR #5002, opened 2026-08-28
Add system test for sorting on rank features

arnej27959:
Covers the new sort-features rank profile block and the corresponding sortspec=feature(<name>) / YQL order by [{"function": "feature"}] query syntax:

 - single-level sorting, including that feature sorting defaults to ascending while attribute sorting takes its default from the schema
 - multi-level sorting mixing feature, attribute and [rank] levels, plus the same feature used on two levels
 - relevance stays 0 when sorting by a feature alone, and comes back when [rank] is part of the sort spec, also for a profile with a second phase
 - sort-features are inherited, and an empty block clears them
 - errors: feature not declared sortable, profile without sort-features, inherited sort-features cleared, illegal feature name, missing() combined with feature(), and a streaming schema


hmusum (approved):
Great!

radu-gheorghe (approved):
Great! Thanks!

PR #5004, opened 2026-08-28
system test select struct fields in document summary

arnej27959:
please have a quick look at some of the schema blocks / json outputs here to verify that this works like we agreed upon.


hmusum (dismissed):

bratseth:
I think we at the end agreed that we should disallow selecting just the key or value in maps so that maps are always rendered as a json object where this can only subtract some of the fields when the value is a struct, and users don't need to know about the special fields 'key' and 'value'? Otherwise LGTM.

arnej27959:
> I think we at the end agreed that we should disallow selecting just the key or value in maps so that maps are always rendered as a json object where this can only subtract some of the fields when the value is a struct, and users don't need to know about the special fields 'key' and 'value'? Otherwise LGTM.

yes - I've been back and forth on this; to be as conservative as possible I'm currently enforcing that you have to select the "key" field for map of structs, instead of having it implicit; it's easy to relax that (making key implicitly-selected) later.

bratseth:
Ok, but taking away the support for getting a different syntax when values is not selected seems harder?

arnej27959:
> Ok, but taking away the support for getting a different syntax when values is not selected seems harder?

yes; the different syntax should be only for map of primitives though.  not sure if we want to allow that, or what syntax we should return it with; maybe
```
"mymapfield": {
     "key": [ "a", "b", "c" ]
}
```
would be better?  but if we require key and at least one value, there's no proper subset possible for map<string,string>

In the PR implementing this I left a toggle to make it easy if we want to disallow it:
https://github.com/vespa-engine/vespa/blob/2da526c8a89a140eee96d81e3ab5daa5ecea9a8c/config-model/src/main/java/com/yahoo/schema/processing/SummaryStructFieldSelectValidator.java#L49

bratseth:
I think "key": [ "a", "b", "c" ] would be better because it is different in any case, and that is at least compact.
But is there any evidence this is needed? Afaik the feature request was for filtering the struct fields.
If we want to allow returning the key set instead of the map, maybe it should just be a separate feature ...

arnej27959:
ok, let's wait with that part - removed from these tests, and flipped the flag in
 https://github.com/vespa-engine/vespa/pull/37735

hmusum (approved):

PR #5006, opened 2026-08-31
Update expected md5sum summary.def

hmusum:
Update after changes in https://github.com/vespa-engine/vespa/pull/37735


toregge (approved):
👍

PR #5010, opened 2026-09-04
Extend test for string ranges

boeker:
Now verifies consistent behavior across indexed and streaming search for cased and uncased fields. I also added a test case verifying match features.

Depends on https://github.com/vespa-engine/vespa/pull/37798.


arnej27959 (approved):
👍

PR #5013, opened 2026-09-07
Test cased matching with fast map search

boeker:
The new test cases with fast map search fail at the moment. They _should_ work after fixing the implementation.

Also re-arranged the existing test cases to make things more clear.


johansolbakken (approved):
Great 👍🏻

PR #5015, opened 2026-09-08
Activate fast map search in map performance test

boeker:
Let's see what happens.

The plan is to update the test later to add a separate field for fast map search. Then one can verify that the queries yield the same results with and without fast map search.


johansolbakken (approved):
🤞🏻

PR #5017, opened 2026-09-08
Add dedicated fast-search test case to map performance test

boeker:
Turns off fast search in the existing test case and adds a new test case to the map performance test that runs the benchmarks with (and only with) the `fast-search` option. (Having all cases in one test case would mean that the test case would takes ages to run, so let's use a separate test case for this for now.)

Moreover, before running the benchmark, the test case verifies that queries return the same hits with and without the `fast-search` option.

Let's wait for a few more test runs before merging.

@johansolbakken fyi


arnej27959 (approved):
👍

PR #5019, opened 2026-09-09
Test lexical ranges sent as JSON/select queries

boeker:
Sends every query also as a JSON/select query. Depends on https://github.com/vespa-engine/vespa/pull/37837.


johansolbakken (approved):
👍🏻

PR #5031, opened 2026-09-18
Refactor fast-map-search test a bit

boeker:
Moves things around, reduces reliance on feed files, merges test cases, and removes some test cases (that tested non-fast-map-search functionality).


johansolbakken (approved):
👍🏻 looks great

